PR #33, opened 2018-07-27
added suppresswarnings for SharedTableBuilder docs that exceed LineLe…

msgroi:
- added suppresswarnings for SharedTableBuilder docs that exceed LineLength and the dozen or so places that are violating the lambda indentation rule which we believe is a bug
- turned on checkstyle fail on error

### DIFF
--- a/src/test/java/com/salesforce/dynamodbv2/DocGeneratorRunner.java
+++ b/src/test/java/com/salesforce/dynamodbv2/DocGeneratorRunner.java
@@ -472,7 +472,6 @@ class DocGeneratorRunner {
             return createItem(hashKeyField, value, "someField", "value-" + value);
         }
 
-        @SuppressWarnings("all")
         private String buildTableName(String table, int ordinal) {
             return buildTableName(table + ordinal);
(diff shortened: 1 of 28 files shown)


src/test/java/com/salesforce/dynamodbv2/DocGeneratorRunner.java
```diff
@@ -100,6 +100,7 @@
     /*
      * logger -> account
      */
+    @SuppressWarnings("checkstyle:Indentation")
```
horwitzj: for when someone (we?) look back at these a while from now and wonder exactly why we suppressed warnings, could you also add a comment (perhaps above the @SuppressWarnings?) akin to "// Checkstyle: 'lambda arguments' has incorrect indentation level 16, expected level should be 12." (or something that mentions it was checkstyle's (apparent) lambda foolishness)? (here and in the other lamba-related locations.)

horwitzj:
(replaced by https://github.com/salesforce/mt-dynamo/pull/41)

PR #35, opened 2018-07-27
replace compare usage with equals

horwitzj:


src/test/java/com/salesforce/dynamodbv2/mt/mappers/MtAmazonDynamoDbStreamTestRunner.java
```diff
@@ -133,7 +133,7 @@ void await(int timeoutSeconds) {
     private void assertMtRecord(MtRecord receivedRecord, List<MtRecord> expectedMtRecords) {
         Optional<MtRecord> mtRecordFound = expectedMtRecords
                 .stream()
-                .filter(mtRecord -> new MtRecordComparator().compare(receivedRecord, mtRecord) == 0)
+                .filter(mtRecord -> weakMtRecordEquals(mtRecord, receivedRecord))
```
horwitzj: it seems that this caller doesn't need the full strength of comparison—just equals. (and it seems like there aren't any other callers to MtRecordComparator().compare.)

PR #53, opened 2018-08-08
Isolated tests

msgroi:


busjaeger:
It's a bit too much for me to review in detail, but we've talked about the intent and I am on board with that. Is there a specific area of the change you'd like me to look at?

msgroi:
@busjaeger Forgot to mention that I already incorporated your suggested change to have field variables and an overloaded constructor in DefaultArgumentProvider.  See 9459fe72c154e9c221a984c7d059501cb5d3abeb.  And I plan to also incorporate your withContext() and eachWithContext() suggestions as well.

busjaeger:
Yes, for sure. Thanks for putting this nice test framework in place and for incorporating my changes.

PR #61, opened 2018-08-14
Fix conditional update

msgroi:


src/main/java/com/salesforce/dynamodbv2/mt/mappers/sharedtable/impl/ConditionMapper.java
```diff
@@ -0,0 +1,142 @@
...
+                        + primaryExpression + ", or filterExpression=" + filterExpression)));
+    }
+
+    /*
+     * Finds the value in the right-hand side operand of an expression where the left-hand operator is a given field.
+     */
+    @VisibleForTesting
+    static Optional<String> findVirtualValuePlaceholder(String conditionExpression, String keyFieldName) {
+        String toFind = keyFieldName + " = ";
+        int start = conditionExpression.indexOf(toFind);
```
busjaeger: What happens if they don't use spaces before and after equals sign?

src/main/java/com/salesforce/dynamodbv2/mt/mappers/sharedtable/impl/ConditionMapper.java
```diff
@@ -0,0 +1,142 @@
...
+        if (start == -1) {
```
msgroi: Jeremy: what happens if there's more than one reference found?

src/main/java/com/salesforce/dynamodbv2/mt/mappers/sharedtable/impl/ConditionMapper.java
```diff
@@ -0,0 +1,180 @@
...
+import java.util.stream.Stream;
+
+/**
+ * Applies mapping and prefixing to condition query and conditional update expressions.
+ *
+ * @author msgroi
+ */
+class ConditionMapper {
+
+    static final String NAME_PLACEHOLDER = "#___name___";
```
msgroi: Jeremy: make it such that this name is not reserved.

msgroi:
Reviewed in person with Jeremy.

PR #72, opened 2018-08-20
added default withContext() impl on MtAmazonDynamoDbContextProvider

msgroi:


msgroi:
Jeremy's taking care of this in his PR.

PR #99, opened 2018-09-11
changed getCreateTableRequest method on CreateTableRequestFactory to …

msgroi:
…return an Optional


msgroi:
Reviewed offline with @horwitzj

PR #107, opened 2018-09-13
Fix checkstyle violations

msgroi:


busjaeger:
Thanks for fixing these violations. Would be great if you or Jeremy could show me how to set up my IDE to apply the right formatting and flag violations, so I can avoid introducing errors going forward.

msgroi:
Replaced this PR with https://github.com/salesforce/mt-dynamo/pull/108

msgroi:
Yeah @busjaeger ... you exposed a couple of problems ... 1) we never documented "here's how to set up Intellij to do checkstyle checks in your IDE and here's preferences you can set to minimize introducing violations" and 2) the travis CI job doesn't run the checkstyle check into it gets into master.  @horwitzj FYI.

PR #109, opened 2018-09-13
travis ci job now runs checkstyle check

msgroi:


msgroi:
Reviewed with Jeremy offline.

PR #110, opened 2018-09-14
Streams implementation for shared table

busjaeger:


src/main/java/com/salesforce/dynamodbv2/mt/mappers/sharedtable/SharedTableBuilder.java
```diff
@@ -100,7 +100,6 @@
 public class SharedTableBuilder extends SharedTableCustomDynamicBuilder {
 
     private List<CreateTableRequest> createTableRequests;
-    private Boolean precreateTables;
```
busjaeger: Moved into base class

src/main/java/com/salesforce/dynamodbv2/mt/mappers/sharedtable/impl/MtAmazonDynamoDbBySharedTable.java
```diff
@@ -118,6 +118,15 @@ public MtAmazonDynamoDbBySharedTable(String name,
...
     }
 
+    List<CreateTableRequest> getSharedTables() {
+        return tableMappingFactory.getCreateTableRequestFactory().getPhysicalTables();
+    }
+
+    @Override
+    protected MtAmazonDynamoDbContextProvider getMtContext() {
+        return super.getMtContext();
+    }
```
busjaeger: Need access to context in MtStreams. Base class declares method as protected in different package, so had to override to make it visible.

src/main/java/com/salesforce/dynamodbv2/mt/mappers/sharedtable/impl/MtAmazonDynamoDbStreamsBySharedTable.java
```diff
@@ -0,0 +1,67 @@
...
+                "listStreams for shared table currently does not support calling with tenant context");
+        checkArgument(listStreamsRequest.getTableName() == null,
+                "listStreams for shared table currently does not support filtering by table name");
+
+        ListStreamsResult result = super.listStreams(listStreamsRequest);
+        Set<String> sharedTables =
+                mtDynamoDb.getSharedTables().stream().map(CreateTableRequest::getTableName).collect(toSet());
+        return result.withStreams(result.getStreams().stream()
+                .filter(stream -> sharedTables.contains(stream.getTableName()))
+                .collect(toList()));
```
busjaeger: Note: the response will contain the physical table name.

src/main/java/com/salesforce/dynamodbv2/mt/mappers/sharedtable/impl/MtAmazonDynamoDbStreamsBySharedTable.java
```diff
@@ -0,0 +1,67 @@
...
+    /**
+     * Returns streams associated with the corresponding MT shared table instance.
+     *
+     * @param listStreamsRequest Stream request. Currently doesn't support filtering by table.
+     * @return Result.
+     */
+    @Override
+    public ListStreamsResult listStreams(ListStreamsRequest listStreamsRequest) {
+        checkArgument(mtDynamoDb.getMtContext().getContext() == null,
+                "listStreams for shared table currently does not support calling with tenant context");
```
busjaeger: I am hoping we can get away without implementing the context-aware variant of this method that would return one stream result per virtual table (where multiple stream results have the same streamArn), since those clients would likely ask for the stream arn via the table directly. This method will likely be used for KCL to start one worker per shared table.

busjaeger: ... maybe I should put that in the javadoc

src/main/java/com/salesforce/dynamodbv2/mt/mappers/sharedtable/CreateTableRequestFactory.java
```diff
@@ -31,8 +31,6 @@
     /*
      * Returns a list of CreateTableRequests that will be created when the factory is initialized.
      */
-    default List<CreateTableRequest> precreateTables() {
-        return new ArrayList<>();
-    }
+    List<CreateTableRequest> getPhysicalTables();
```
busjaeger: It may make sense to store the list of physical tables in MtAmazonDynamoDbBySharedTable and either turn this back into a functional interface (with only the table mapping method) or make table mapping the responsibility of the TableMapper and delete this interface.

src/main/java/com/salesforce/dynamodbv2/mt/mappers/sharedtable/impl/MtAmazonDynamoDbStreamsBySharedTable.java
```diff
@@ -0,0 +1,67 @@
...
+        Set<String> sharedTables =
+                mtDynamoDb.getSharedTables().stream().map(CreateTableRequest::getTableName).collect(toSet());
+        return result.withStreams(result.getStreams().stream()
+                .filter(stream -> sharedTables.contains(stream.getTableName()))
+                .collect(toList()));
+    }
+
+    @Override
+    public GetRecordsResult getRecords(GetRecordsRequest request) {
+        throw new UnsupportedOperationException();
```
busjaeger: I'll implement this method in a separate PR.

src/test/java/com/salesforce/dynamodbv2/mt/mappers/sharedtable/impl/ItemMapperTest.java
```diff
@@ -31,7 +30,7 @@
     private static final String PREFIX = "PREFIX-";
     private static final ItemMapper SUT = new ItemMapper(new TableMapping(new DynamoTableDescriptionImpl(
         CreateTableRequestBuilder.builder().withTableKeySchema("virtualhk", S).build()),
-        virtualTableDescription1 -> Optional.of(new DynamoTableDescriptionImpl(CreateTableRequestBuilder.builder()
+            new SingletonCreateTableRequestFactory(new DynamoTableDescriptionImpl(CreateTableRequestBuilder.builder()
```
busjaeger: Had to introduce this class, since CreateTableRequestFactory is no longer a functional interface (see comments above).

PR #111, opened 2018-09-17
GetRecords implementation for shared table streams

busjaeger:


src/main/java/com/salesforce/dynamodbv2/mt/context/impl/MtAmazonDynamoDbContextProviderImpl.java
```diff
@@ -29,12 +29,7 @@ public void setContext(String tenantId) {
 
     @Override
     public String getContext() {
-        String value = getContextMap().get(CONTEXT_KEY);
-        if (value == null || value.trim().isEmpty()) {
-            throw new IllegalStateException("no context available");
-        } else {
-            return value;
-        }
+        return getContextMap().get(CONTEXT_KEY);
```
msgroi: is it ok to return null?

src/main/java/com/salesforce/dynamodbv2/mt/mappers/MtAmazonDynamoDbStreams.java
```diff
@@ -25,6 +26,9 @@ static MtAmazonDynamoDbStreams createFromDynamo(AmazonDynamoDB dynamoDb, AmazonD
         }
 
         if (dynamoDb instanceof MtAmazonDynamoDbBySharedTable) {
+            AmazonDynamoDBStreams streams = dynamoDbStreams instanceof CachingAmazonDynamoDbStreams
```
msgroi: Seems like caching would never be turned on here.

src/main/java/com/salesforce/dynamodbv2/mt/mappers/sharedtable/impl/CachingAmazonDynamoDbStreams.java
```diff
@@ -87,36 +87,39 @@
         private long getRecordsLimitExceededBackoffInMillis =
                 DEFAULT_GET_RECORDS_LIMIT_EXCEEDED_BACKOFF_IN_MILLIS;
 
-        Builder(AmazonDynamoDBStreams amazonDynamoDbStreams) {
+        public Builder(AmazonDynamoDBStreams amazonDynamoDbStreams) {
```
msgroi: This triggers annoying "access can be package-private" warnings in Intellij.  I generally like having that warning enabled but in this case we don't want it.

src/main/java/com/salesforce/dynamodbv2/mt/context/MtAmazonDynamoDbContextProvider.java
```diff
@@ -37,4 +39,25 @@ default void withContext(String tenantId, Runnable runnable) {
...
+     * original tenantId, and returns the result of calling the function.
+     *
+     * @param tenantId Context tenantId to use when calling the function.
+     * @param function Function to call within tenant context.
+     * @param t        Parameter to function.
+     * @param <T>      Input type of function.
+     * @param <R>      Output type of function.
+     * @return
+     */
+    default <T, R> R withContext(String tenantId, Function<T, R> function, T t) {
```
msgroi: Nice

src/main/java/com/salesforce/dynamodbv2/mt/mappers/sharedtable/impl/MtAmazonDynamoDbStreamsBySharedTable.java
```diff
@@ -41,22 +88,132 @@ public MtAmazonDynamoDbStreamsBySharedTable(AmazonDynamoDBStreams amazonDynamoDb
...
+                        .withOldImage(itemMapper.reverse(streamRecord.getOldImage()))
+                        .withSequenceNumber(streamRecord.getSequenceNumber())
+                        .withStreamViewType(streamRecord.getStreamViewType())
+                        .withApproximateCreationDateTime(streamRecord.getApproximateCreationDateTime())
+                        .withSizeBytes(streamRecord.getSizeBytes()));
+    }
+
+    private boolean matchesContext(MtRecord mtRecord) {
+        String context = mtDynamoDb.getMtContext().getContext();
+        return context == null || context.equals(mtRecord.getContext());
```
msgroi: I see, you're treating no context as "all" contexts.  This puts more responsibility on higher level clients to make sure there is a context provided where appropriate.

src/main/java/com/salesforce/dynamodbv2/mt/mappers/sharedtable/impl/MtAmazonDynamoDbStreamsBySharedTable.java
```diff
@@ -41,22 +88,132 @@ public MtAmazonDynamoDbStreamsBySharedTable(AmazonDynamoDBStreams amazonDynamoDb
...
+            LOG.debug("listStreams request={}", listStreamsRequest);
+        }
+
         checkArgument(mtDynamoDb.getMtContext().getContext() == null,
                 "listStreams for shared table currently does not support calling with tenant context");
         checkArgument(listStreamsRequest.getTableName() == null,
                 "listStreams for shared table currently does not support filtering by table name");
 
+        // filter to mt tables
         ListStreamsResult result = super.listStreams(listStreamsRequest);
```
msgroi: So to summarize, this code gets all streams, then gets the physical tables associated with this SharedTables instance, and filters down to the list of streams for those shared tables only?

src/main/java/com/salesforce/dynamodbv2/mt/context/impl/MtAmazonDynamoDbContextProviderImpl.java
busjaeger: Will change this based on our discussion.

src/main/java/com/salesforce/dynamodbv2/mt/mappers/MtAmazonDynamoDbStreams.java
busjaeger: I also noticed that when prepping for the demo :). Impressive that you caught this just by inspecting the code!

src/main/java/com/salesforce/dynamodbv2/mt/mappers/sharedtable/impl/CachingAmazonDynamoDbStreams.java
busjaeger: I had to make it public so that MtAmazonDynamoDbStreams (which is in another package) is able to access it, so hopefully we won't get warnings here. But I do see the warnings for the with* methods, since it doesn't use those.

src/main/java/com/salesforce/dynamodbv2/mt/mappers/sharedtable/impl/MtAmazonDynamoDbStreamsBySharedTable.java
busjaeger: Yes, exactly.

We could enable this method to work in the context of a tenant in the future.

src/main/java/com/salesforce/dynamodbv2/mt/mappers/sharedtable/impl/MtAmazonDynamoDbStreamsBySharedTable.java
busjaeger: Yes. Will change this as discussed.

src/main/java/com/salesforce/dynamodbv2/mt/mappers/sharedtable/impl/CachingAmazonDynamoDbStreams.java
msgroi: Yeah, this was more of a complaint about Intellij than a request to make a change.  :-)

src/test/java/com/salesforce/dynamodbv2/mt/mappers/sharedtable/impl/MtAmazonDynamoDbStreamsBySharedTableTest.java
```diff
@@ -214,37 +213,43 @@ void testRecords() {
...
-            MT_CONTEXT.withContext(TENANTS[1], () -> assertGetRecords(mtDynamoDbStreams, iterator, expected3,
-                    expected4));
+            MT_CONTEXT.withContext(TENANTS[0],
+                () -> assertGetRecords(mtDynamoDbStreams, iterator, expected1, expected2));
+            MT_CONTEXT.withContext(TENANTS[1],
+                () -> assertGetRecords(mtDynamoDbStreams, iterator, expected3, expected4));
+
+            // once to fetch records, once per getRecords (3) to find empty range
+            assertEquals(4, dynamoDbStreams.getRecordsCount);
+            assertEquals(1, dynamoDbStreams.getShardIteratorCount);
```
busjaeger: Mostly reformatting. This is the actual change in this class (verifying number of cache misses).

PR #112, opened 2018-09-18
removed 3 unnecessary @SuppressWarning's

msgroi:


msgroi:
Review offline with Jeremy.

PR #116, opened 2018-09-19
force javadoc compliance at earlier phase

msgroi:


Makefile
```diff
@@ -1,10 +1,11 @@
 build:
-	mvn test-compile -DskipTests=true -Dmaven.javadoc.skip=true -B -V
```
msgroi: -Dmaven.javadoc.skip=true didn't seem to be doing anything that I could observe

msgroi:
reviewed with jeremy offline

PR #131, opened 2018-09-25
GT support

horwitzj:


src/test/java/com/salesforce/dynamodbv2/testsupport/ArgumentBuilder.java
```diff
@@ -324,7 +392,7 @@ public void forEachOrgContext(Consumer<String> consumer) {
         @Override
         public String toString() {
             return amazonDynamoDb.getClass().getSimpleName()
-                + ", orgs=" + orgs + '}'
```
horwitzj: is this right?

msgroi: looks like a good change to me

src/test/java/com/salesforce/dynamodbv2/testsupport/DefaultTestSetup.java
```diff
@@ -84,26 +94,87 @@ public void setupTableData(AmazonDynamoDB amazonDynamoDb,
...
+                        + " encountered");
+            }
         }
     }
 
+    /**
+     * Gets the {@code ScalarAttributeType} associated with the one key in the key schema with {@code KeyType}
+     * {@code keyType}.
+     */
+    private Optional<ScalarAttributeType> getScalarAttributeType(CreateTableRequest createTableRequest,
```
msgroi: This is the kind of thing you might want a test for.

PR #142, opened 2018-09-28
Filter records for disabled streams from shared stream

msgroi:


src/main/java/com/salesforce/dynamodbv2/mt/mappers/sharedtable/impl/MtAmazonDynamoDbStreamsBySharedTable.java
```diff
@@ -81,6 +92,9 @@ private MtRecord mapRecord(Function<Map<String, AttributeValue>, FieldValue> fie
         FieldValue fieldValue = fieldValueFunction.apply(record.getDynamodb().getKeys());
         MtAmazonDynamoDbContextProvider mtContext = mtDynamoDb.getMtContext();
         // execute in record tenant context to get table mapping
+        System.out.println(
+            fieldValue.getMtContext() + " " + fieldValue.getTableIndex() + " " + fieldValue.getUnqualifiedValue());
+
```
busjaeger: probably want to take this out (sorry, forgot to remove it)

src/test/java/com/salesforce/dynamodbv2/mt/mappers/MtAmazonDynamoDbStreamsBaseTest.java
```diff
@@ -277,4 +287,59 @@ void testTableStream(MtAmazonDynamoDbBase mtDynamoDb) {
...
+                String iterator = getShardIterator(mtDynamoDbStreams, streamArn).get();
+                assertGetRecords(mtDynamoDbStreams, iterator, streamsEnabledRecord);
+            });
+
+            // verify that either the streamArn is null(for ByTable or ByAccount) or the iterator returns no records
+            MT_CONTEXT.withContext(tenant, () -> {
+                String streamArn = mtDynamoDb.describeTable(tableWithStreamsDisabled).getTable().getLatestStreamArn();
+                if (streamArn != null) {
+                    // for byIndex, the streamArn will not be null, but it should return no records
+                    String iterator = getShardIterator(mtDynamoDbStreams, streamArn).get();
```
msgroi: So this test currently fails when it tests ByIndex, because it gets records from the other table that it is co-located whose virtual table has streams enabled.  Seems like the test is right and the implementation is wrong no?

src/main/java/com/salesforce/dynamodbv2/mt/mappers/sharedtable/impl/MtAmazonDynamoDbStreamsBySharedTable.java
msgroi: removed https://github.com/salesforce/mt-dynamo/pull/142/commits/81a062ac8d70516e5729036853ce5dbd380fdf5c

src/test/java/com/salesforce/dynamodbv2/mt/mappers/MtAmazonDynamoDbStreamsBaseTest.java
```diff
@@ -277,4 +287,59 @@ void testTableStream(MtAmazonDynamoDbBase mtDynamoDb) {
...
+                newCreateTableRequest(tableWithStreamsDisabled)
+                    .withStreamSpecification(new StreamSpecification()
+                        .withStreamEnabled(false)
+                    );
+            MT_CONTEXT.withContext(tenant, () -> mtDynamoDb.createTable(createTableRequestStreamsDisabled));
+
+            // put an item in each table
+            final MtRecord streamsEnabledRecord = putTestItem(mtDynamoDb, tableWithStreamsEnabled, tenant, 1);
+            putTestItem(mtDynamoDb, tableWithStreamsDisabled, tenant, 2);
+
```
busjaeger: Would be good to add a variant that queries the stream without context.

src/test/java/com/salesforce/dynamodbv2/mt/mappers/MtAmazonDynamoDbStreamsBaseTest.java
```diff
@@ -277,4 +287,59 @@ void testTableStream(MtAmazonDynamoDbBase mtDynamoDb) {
...
+            MT_CONTEXT.withContext(tenant, () -> {
+                String streamArn = mtDynamoDb.describeTable(tableWithStreamsEnabled).getTable().getLatestStreamArn();
+                assertNotNull(streamArn);
+                String iterator = getShardIterator(mtDynamoDbStreams, streamArn).get();
+                assertGetRecords(mtDynamoDbStreams, iterator, streamsEnabledRecord);
+            });
+
+            // verify that either the streamArn is null(for ByTable or ByAccount) or the iterator returns no records
+            MT_CONTEXT.withContext(tenant, () -> {
+                String streamArn = mtDynamoDb.describeTable(tableWithStreamsDisabled).getTable().getLatestStreamArn();
```
busjaeger: Should the streamArn returned here be null if queried in the context of a tenant for which the arn is disabled?

src/test/java/com/salesforce/dynamodbv2/mt/mappers/MtAmazonDynamoDbStreamsBaseTest.java
busjaeger: Agree, that's a bug. The implementation is only filtering out records for other tenants, but not records for other virtual tables. We will probably need to give each virtual table it's own virtual arn, so that getShardIterator/getRecords knows what to return.

busjaeger: https://docs.aws.amazon.com/general/latest/gr/aws-arns-and-namespaces.html

src/test/java/com/salesforce/dynamodbv2/mt/mappers/MtAmazonDynamoDbStreamsBaseTest.java
msgroi: yes, will do

src/test/java/com/salesforce/dynamodbv2/mt/mappers/MtAmazonDynamoDbStreamsBaseTest.java
msgroi: It seems like with a native AmazonDynamoDB instance, if you describe a table with streaming disabled, then getLatestStreamArn() returns null.  So if we want this method to behave consistently with native behavior(and ByTable), then yes, it should always return null at the line you're referring to.

src/test/java/com/salesforce/dynamodbv2/mt/mappers/MtAmazonDynamoDbStreamsBaseTest.java
```diff
@@ -191,6 +193,14 @@ private static void assertGetRecords(MtAmazonDynamoDbStreams streams, Collection
 
             AmazonDynamoDB dynamoDb = AmazonDynamoDbLocal.getAmazonDynamoDbLocal();
 
+            boolean loggingEnabled = false;
```
horwitzj: is this (196–203) debugging code?

busjaeger:
@msgroi the latest commit fixes the test case you added and is ready for review.

busjaeger:
@horwitzj would be great if you could take a look as well.

src/main/java/com/salesforce/dynamodbv2/mt/util/StreamArn.java
```diff
@@ -0,0 +1,211 @@
...
+
+        @Override
+        public String toString() {
+            return String.format(VIRTUAL_FORMAT, super.toString(), context, mtTableName);
+        }
+    }
+
+    private static final String FORMAT = "arn:%s:%s:%s:%s:table/%s/stream/%s";
+
+    private static final Pattern ARN_PATTERN = Pattern.compile("arn:(?<partition>[^:]+):(?<service>[^:]+):"
```
horwitzj: will this be (too) slow?

busjaeger: I replaced it with manual parsing in latest commit.

src/main/java/com/salesforce/dynamodbv2/mt/mappers/MtAmazonDynamoDbStreamsBase.java
```diff
@@ -166,15 +139,17 @@ public GetRecordsResult getRecords(GetRecordsRequest request) {
             LOG.debug("getRecords request={}", request);
         }
 
-        ShardIterator iterator = ShardIterator.parse(request.getShardIterator());
+        ShardIterator iterator = ShardIterator.fromString(request.getShardIterator());
+        String arn = iterator.getArn();
+        StreamArn streamArn = parse(arn);
```
msgroi: If we're worried about performance we could memoize parsing of arn's.  But I'm generally in favor of not prematurely optimizing this especially given that I'm not even sure regionMatches is going to be a lot slower than doing it ourselves.  So, in conclusion, I think it's fine to merge as-is.

msgroi:
LGTM

PR #148, opened 2018-10-03
Caching streams improvements

busjaeger:


src/main/java/com/salesforce/dynamodbv2/mt/util/CachingAmazonDynamoDbStreams.java
```diff
@@ -0,0 +1,964 @@
...
+            return new IteratorPosition(streamArn, shardId, sequenceNumber.add(ONE));
+        }
+
+        /**
+         * Returns the position immediately following the last record in the given result.
+         *
+         * @param result Result to position after. Record list must not be empty.
+         * @return IteratorPosition immediately following the last record sequence number in the result.
+         */
+        IteratorPosition next(GetRecordsResult result) {
```
msgroi: Maybe rename this or do something to distinguish from next() which is strictly dependent on the sequence.  Same comment on nextShardIterator().

PR #165, opened 2018-10-16
Separate cache for shard iterators to invalidate expired

busjaeger:


busjaeger:
Another attempt at getting shard iterator caching right :). Basic idea:
- LATEST and TRIM_HORIZON are never cached (as before)
- cached GetRecordResults don't cache underlying DynamoDB iterator (only logical pointer)
- AT_SEQUENCE_NUMBER and AFTER_SEQUENCE_NUMBER are cached (both GetShardIterator result and loaded GetRecordResults)
- if we catch an expired iterator exception for a cached iterator --> clear cache and try again

src/main/java/com/salesforce/dynamodbv2/mt/util/CachingAmazonDynamoDbStreams.java
```diff
@@ -414,51 +431,46 @@ private CachingShardIterator(
...
         }
 
         /**
-         * Returns underlying streams iterator, loading it if present.
+         * Returns a new iterator with the given dynamoDbIterator.
          *
-         * @param streams Streams instance to use for loading iterator if needed.
-         * @return DynamoDB iterator.
+         * @param dynamoDbIterator DynamoDb iterator.
+         * @return New iterator.
```
horwitzj: "New" -> "a new"

PR #168, opened 2018-10-18
Fixed NPE in SharedTable when performing an update request on a GSI h…

msgroi:
Fixed this error ...

https://gist.github.com/msgroi/e330717e8ec86e8ba93c699ee3c97073

On a request that updated multiple fields, including a GSI HK field, where the GSI HK field wasn't the last field in the set x = y pattern.

### DIFF
--- a/src/main/java/com/salesforce/dynamodbv2/mt/mappers/sharedtable/impl/ConditionMapper.java
+++ b/src/main/java/com/salesforce/dynamodbv2/mt/mappers/sharedtable/impl/ConditionMapper.java
@@ -170,17 +170,19 @@ class ConditionMapper {
         if (conditionExpression == null) {
             return Optional.empty();
         }
-        String toFind = keyFieldName + " = ";
-        int start = conditionExpression.indexOf(toFind);
+        // conditionExpression = "set #ue1 = :ue1, #ue2 = :ue2"
+        // TODO add support for non-EQ operators
+        String toFind = keyFieldName + " = "; // "#ue1 = "
+        int start = conditionExpression.indexOf(toFind); // 4
         if (start == -1) {
             return Optional.empty();
         }
 
-        String remainder = conditionExpression.substring(start + toFind.length());
+        String remainder = conditionExpression.substring(start + toFind.length()); // ":ue1, #ue2 = :ue2"
 
-        int end = indexOfUsingRegex(SPACE_OR_PIPE, remainder);
+        int end = indexOfUsingRegex(SPACE_OR_PIPE, remainder); // 4
 
-        return Optional.of(end == -1 ? remainder : remainder.substring(0, end));
+        return Optional.of(end == -1 ? remainder : remainder.substring(0, end)); // ":ue1"
     }
 
     private static int indexOfUsingRegex(String regex, String str) {
--- a/src/main/java/com/salesforce/dynamodbv2/mt/mappers/sharedtable/impl/ConditionMapper.java
+++ b/src/main/java/com/salesforce/dynamodbv2/mt/mappers/sharedtable/impl/ConditionMapper.java
@@ -170,24 +170,11 @@ class ConditionMapper {
         if (conditionExpression == null) {
             return Optional.empty();
         }
-        // conditionExpression = "set #ue1 = :ue1, #ue2 = :ue2"
+
         // TODO add support for non-EQ operators
-        String toFind = keyFieldName + " = "; // "#ue1 = "
-        int start = conditionExpression.indexOf(toFind); // 4
-        if (start == -1) {
-            return Optional.empty();
-        }
-
-        String remainder = conditionExpression.substring(start + toFind.length()); // ":ue1, #ue2 = :ue2"
-
-        int end = indexOfUsingRegex(SPACE_OR_PIPE, remainder); // 4
-
-        return Optional.of(end == -1 ? remainder : remainder.substring(0, end)); // ":ue1"
-    }
-
-    private static int indexOfUsingRegex(String regex, String str) {
-        Matcher matcher = Pattern.compile(regex).matcher(str);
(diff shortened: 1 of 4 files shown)


src/main/java/com/salesforce/dynamodbv2/mt/mappers/sharedtable/impl/ConditionMapper.java
```diff
@@ -172,9 +174,17 @@ static String getNextFieldPlaceholder(Map<String, String> expressionAttributeNam
         if (start == -1) {
             return Optional.empty();
         }
-        int end = conditionExpression.indexOf(" ", start + toFind.length()); // -1
-        return Optional.of(conditionExpression.substring(start + toFind.length(),
-            end == -1 ? conditionExpression.length() : end)); // ":value" // TODO add support for non-EQ operators
+
```
horwitzj: re line 165: which comments?

src/main/java/com/salesforce/dynamodbv2/mt/mappers/sharedtable/impl/ConditionMapper.java
```diff
@@ -172,9 +174,17 @@ static String getNextFieldPlaceholder(Map<String, String> expressionAttributeNam
...
+        String remainder = conditionExpression.substring(start + toFind.length());
+
+        int end = indexOfUsingRegex(" |,", remainder);
+
+        return Optional.of(end == -1 ? remainder : remainder.substring(0, end));
+    }
+
+    private static int indexOfUsingRegex(String regex, String str) {
```
horwitzj: perhaps inline and make Pattern.compile(" |") a constant (rather than rebuilding it every time?)

msgroi: fixed in 7e766ac32745b44342100ca2002b70280815f90f

src/main/java/com/salesforce/dynamodbv2/mt/mappers/sharedtable/impl/ConditionMapper.java
horwitzj: why not replace this whole method with

        if (conditionExpression == null) {
            return Optional.empty();
        }

        final Pattern p = Pattern.compile(keyFieldName + " = ([^ ,]*)");
        final Matcher m = p.matcher(conditionExpression);
        return m.find() ? Optional.of(m.group(1)) : Optional.empty();

msgroi: eek ... unintentionally removed comments, will fix

msgroi: fixed in ae504908b0fd0ce9ae5b2b5b580bdf0612ccfbcc and also added similar cueing that I had in other methods in this case

busjaeger:
What was the original bug?

src/main/java/com/salesforce/dynamodbv2/mt/mappers/sharedtable/impl/ConditionMapper.java
msgroi: alrighty then, fixed ... 541fcfbe15405c70d04d194837df5f7045da97fa

src/main/java/com/salesforce/dynamodbv2/mt/mappers/sharedtable/impl/ConditionMapper.java
```diff
@@ -17,6 +19,7 @@
 class ConditionMapper {
 
     static final String NAME_PLACEHOLDER = "#___name___";
+    private static final String SPACE_OR_PIPE = " |,";
```
horwitzj: no longer needed(?)

msgroi:
@busjaeger The original bug was that you'd get a NPE on a request that updated multiple fields, including a GSI HK field, where the GSI HK field wasn't the last field in the set x = y pattern, e.g. set gsihk = foo, somefield = bar.  I guess I should have created a test that demonstrates reproduced this in UpdateTest?

src/main/java/com/salesforce/dynamodbv2/mt/mappers/sharedtable/impl/ConditionMapper.java
```diff
@@ -167,14 +170,11 @@ static String getNextFieldPlaceholder(Map<String, String> expressionAttributeNam
...
-            return Optional.empty();
-        }
-        int end = conditionExpression.indexOf(" ", start + toFind.length()); // -1
-        return Optional.of(conditionExpression.substring(start + toFind.length(),
-            end == -1 ? conditionExpression.length() : end)); // ":value" // TODO add support for non-EQ operators
+
+        // TODO add support for non-EQ operators
+        final Pattern p = Pattern.compile(keyFieldName + " = ([^ ,]*)");
+        final Matcher m = p.matcher(conditionExpression);
+        return m.find() ? Optional.of(m.group(1)) : Optional.empty();
```
horwitzj: takes the first match found—if there happen to be more than one match, is that what we want here? (do we want to reject (as malformed, e.g.) if there is more than one match on keyFieldName?)

msgroi: The goal is to find the first delimiter in the string and pull the value out that comes before that delimiter.  For a "set" clause, that could be either a space, a comma, or nothing.

msgroi:
@busjaeger I added a test that demonstrates the problem ... f801411fee6699a4409cf6b4bbf800957681954a ... note that it's t's @Disabled because it still fails because of the bug fixed in PR#169.  I will enable it in PR#169 once this PR is merged

PR #197, opened 2018-11-20
Advance trim horizon iterator even if no records found

busjaeger:


src/main/java/com/salesforce/dynamodbv2/mt/util/CachingAmazonDynamoDbStreams.java
```diff
@@ -811,12 +811,13 @@ private GetRecordsResult getRecords(CachingShardIterator iterator) {
...
-                        iteratorCache.put(iterator, loadedNextIterator);
-                        return iterator;
-                    });
+                CachingShardIterator nextIterator;
+                if (iterator.getDynamoDbIterator().isPresent()) {
+                    nextIterator = iterator.withDynamoDbIterator(loadedNextIterator);
+                } else {
+                    iteratorCache.put(iterator, loadedNextIterator);
+                    nextIterator = iterator;
+                }
```
busjaeger: The previous version of the code was not returning the new loaded iterator which caused streams to get stuck at `TRIM_HORIZON`.

busjaeger:
@horwitzj would be great if you could review while Mike is out. Happy to jump on a hangout tomorrow and explain the issue if that would be helpful.

PR #206, opened 2018-11-27
Empty string as base context instead of null

horwitzj:


src/main/java/com/salesforce/dynamodbv2/mt/context/impl/MtAmazonDynamoDbContextProviderImpl.java
```diff
@@ -28,9 +28,9 @@ public void setContext(String tenantId) {
     }
 
     @Override
-    public Optional<String> getContextOpt() {
-        String value = getContextMap().get(CONTEXT_KEY);
-        return value == null || value.trim().isEmpty() ? Optional.empty() : Optional.of(value);
+    public String getContext() {
```
busjaeger: Before this change we only supported “empty” context for scans and records. Now we support it for all methods. Is that correct? If so, would be good to run this by Mike, since he had some feedback on the original PR that introduced the optional.

src/main/java/com/salesforce/dynamodbv2/mt/context/MtAmazonDynamoDbContextProvider.java
```diff
@@ -18,11 +17,7 @@
 @FunctionalInterface
 public interface MtAmazonDynamoDbContextProvider {
 
-    Optional<String> getContextOpt();
-
-    default String getContext() {
-        return getContextOpt().orElseThrow(IllegalStateException::new);
-    }
+    String getContext();
```
busjaeger: There is a pretty significant behavior change here, which is that it's now possible to put/get items in system context which was not possible before. Do we actually want that behavior?

PR #217, opened 2018-12-01
Rename & refactor MtAmazonDynamoDbContextProviderImpl

seratch:
This pull request suggests two changes to MtAmazonDynamoDbContextProviderImpl as follows.

## Rename the impl class not to be confusing

The setter method just updates ThreadLocal value. That means that if a user use the context provider in different threads, it doesn't work as expected. The behavior is surprising and not intuitive. Thus, I suggest to rename it to let users easily understand the characteristics from the name.

## Refactor the internal ThreadLocal value

Generally speaking, a ThreadLocal value can be a static field because the value itself is not shared among threads. In my understanding, doing like this is a common practice. Also, I found that we can specify not `Object` but `Map<String, String>` as the generic type because we won't put other type values into it.

### DIFF
--- a/src/main/java/com/salesforce/dynamodbv2/mt/context/impl/MtAmazonDynamoDbContextProviderThreadLocalImpl.java
+++ b/src/main/java/com/salesforce/dynamodbv2/mt/context/impl/MtAmazonDynamoDbContextProviderThreadLocalImpl.java
@@ -16,11 +16,11 @@ import java.util.Map;
  *
  * @author msgroi
  */
-public class MtAmazonDynamoDbContextProviderImpl implements MtAmazonDynamoDbContextProvider {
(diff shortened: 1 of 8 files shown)


seratch:
```
[ERROR] /home/travis/build/salesforce/mt-dynamo/src/test/java/com/salesforce/dynamodbv2/testsupport/DocGeneratorRunner.java:106: Line is longer than 120 characters (found 123). [LineLength]
[ERROR] /home/travis/build/salesforce/mt-dynamo/src/test/java/com/salesforce/dynamodbv2/testsupport/ArgumentBuilder.java:74: Line is longer than 120 characters (found 122). [LineLength]
```

OK, I will fix them later. Sorry, I didn't run the build on my local machine after renaming the class.

msgroi:
I like both changes.  The rename helps makes it obvious that there may be side effects to using it.   And making the instance variable static is good.  Thanks so much!

PR #221, opened 2018-12-04
instead of checking *equals* "ing", essentially check ends with "ing"

horwitzj:


src/main/java/com/salesforce/dynamodbv2/mt/admin/AmazonDynamoDbAdminUtils.java
```diff
@@ -120,9 +121,18 @@ private String getTableStatus(String tableName) throws TableInUseException {
         try {
             String status = describeTable(tableName).getTableStatus();
             log.info("table=" + tableName + " is " + status);
-            if (status.toLowerCase().equals("ing")) {
-                throw new TableInUseException(tableName, status);
+            final TableStatus tableStatus = TableStatus.fromValue(status.toUpperCase());
```
seratch: Just in case, calling `#toString(Locale.ENGLISH)` is safer regarding Turkish locale environment.
https://docs.oracle.com/javase/8/docs/api/java/lang/String.html#toUpperCase-java.util.Locale-

PR #228, opened 2018-12-06
converted delimiter from String to char

msgroi:


src/main/java/com/salesforce/dynamodbv2/mt/mappers/sharedtable/impl/FieldPrefixFunction.java
```diff
@@ -24,9 +24,8 @@
 
     private final CompositeStrings compositeStrings;
 
-    FieldPrefixFunction(String delimiter) {
-        Preconditions.checkArgument(delimiter.length() == 1, "only single character delimiters are supported");
-        compositeStrings = new CompositeStrings(delimiter.charAt(0), '\\');
+    FieldPrefixFunction(char delimiter) {
```
horwitzj: `Preconditions` import now unused

msgroi: fixed in dacb52758e9f893fabeaea5d8c5007ad319b1adf

PR #259, opened 2019-01-17
Enable on demand tbl creation

degaydorus:
- added support for creating tables with PAY_PER_REQUEST billing mode type: either through default shared table creation requests or through customized shared table requests.
- bug fixed: table creations were being interrupted due to TableInUse improperly being thrown.  Now TableInUse is thrown when the expected operation does not match the current table status when getTableStatus is called
- added tests for AmazonDynamoDbAdminUtils
- note some of the tests test against remote dynamo due to limitations of local dynamo


src/integration/java/com/salesforce/dynamodbv2/mt/admin/AmazonDynamoDbAdminUtilsIt.java
```diff
@@ -0,0 +1,122 @@
...
+                        .withKeySchema(new KeySchemaElement(HASH_KEY_FIELD, KeyType.HASH),
+                                new KeySchemaElement(INDEX_FIELD, KeyType.RANGE))
+                        .withProjection(new Projection().withProjectionType(ProjectionType.ALL)));
+    }
+
+    @Test
+    void createTableIfNotExistsButTableInUseForCreating() {
+
+        remoteDynamoDB.createTable(getTestCreateTableRequest(fullTableName));
+        remoteUtils.createTableIfNotExists(getTestCreateTableRequest(fullTableName), 10);
```
msgroi: Given the name of the table being used, this test will be subject to race conditions if it is running more than once against a given AWS instance correct?

src/integration/java/com/salesforce/dynamodbv2/mt/admin/AmazonDynamoDbAdminUtilsIt.java
```diff
@@ -0,0 +1,122 @@
...
+
+        // cleanup
+        remoteUtils.deleteTableIfExists(fullTableName, 10, 600);
+    }
```
msgroi: Should the clean up be in @After or @AfterClass?

src/integration/java/com/salesforce/dynamodbv2/mt/admin/AmazonDynamoDbAdminUtilsIt.java
```diff
@@ -0,0 +1,122 @@
+package com.salesforce.dynamodbv2.mt.admin;
```
msgroi: Is there a benefit to putting this in src/integration/java instead of src/test/java?

src/integration/java/com/salesforce/dynamodbv2/mt/admin/AmazonDynamoDbAdminUtilsIt.java
msgroi: Sorry, I now see that you're suffixing with timestamp.  The problem with this is if the test bails, then you leave behind a table that will never be cleaned up by subsequent test executions.  I suggest either ... 1) having the clean up find an purge all tables that start with table prefix(that way those tables eventually get cleaned up), or 2) including the hostname of the machine executing the test and cleaning up afterwards.

src/integration/java/com/salesforce/dynamodbv2/mt/mappers/sharedtable/impl/SharedTableBuilderIt.java
```diff
@@ -0,0 +1,98 @@
...
+
+        assertEquals(BillingMode.PAY_PER_REQUEST.toString(), remoteDynamoDB.describeTable(TABLE_PREFIX + tableName)
+                .getTable().getBillingModeSummary().getBillingMode());
+
+        remoteDynamoDB.deleteTable(TABLE_PREFIX + tableName);
+    }
+
+    @Test
+    @Disabled
+    void testBillingModePayPerRequestWithDefaults() {
```
msgroi: If this is testing the default builder settings then I wouldn't expect to see `.withDefaultProvisionedThroughput(0)`.  If it isn't, then the test should be renamed.

src/test/java/com/salesforce/dynamodbv2/mt/mappers/sharedtable/SharedTableBuilderTest.java
```diff
@@ -0,0 +1,88 @@
...
+import com.amazonaws.services.dynamodbv2.model.ProjectionType;
+import com.amazonaws.services.dynamodbv2.model.ProvisionedThroughput;
+import com.salesforce.dynamodbv2.dynamodblocal.AmazonDynamoDbLocal;
+import com.salesforce.dynamodbv2.mt.context.MtAmazonDynamoDbContextProvider;
+import com.salesforce.dynamodbv2.mt.context.impl.MtAmazonDynamoDbContextProviderThreadLocalImpl;
+import com.salesforce.dynamodbv2.mt.mappers.sharedtable.impl.MtAmazonDynamoDbBySharedTable;
+import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.api.Test;
+
+class SharedTableBuilderTest {
```
msgroi: I'm assuming you're not testing provisionedThroughput=0 because this test won't work against DynamoDBLocal?

src/integration/java/com/salesforce/dynamodbv2/mt/admin/AmazonDynamoDbAdminUtilsIt.java
degaydorus: good point, I will have a "just in case" cleanup step at the end. If each test runs through, they each have a cleanup line depending on the test.

src/integration/java/com/salesforce/dynamodbv2/mt/admin/AmazonDynamoDbAdminUtilsIt.java
degaydorus: I was following the convention of our other projects.  I think it's helpful to have them separate, it's easier to differentiate the two when first looking at the project.

src/integration/java/com/salesforce/dynamodbv2/mt/admin/AmazonDynamoDbAdminUtilsIt.java
degaydorus: moved all cleanup to afterAll

src/integration/java/com/salesforce/dynamodbv2/mt/admin/AmazonDynamoDbAdminUtilsIt.java
degaydorus: ended up moving all cleanup work to afterAll

src/integration/java/com/salesforce/dynamodbv2/mt/mappers/sharedtable/impl/SharedTableBuilderIt.java
degaydorus: It's testing that Pay Per Request billing mode gets set for buildDefaultCreateTableRequests.  Renaming to "testBillingModePayPerRequestIsSetForDefaultCreateTableRequests"

src/test/java/com/salesforce/dynamodbv2/mt/mappers/sharedtable/SharedTableBuilderTest.java
degaydorus: Yes, the tests for provisionedThroughput=0 are in /src/integration/java/../SharedTableBuilderIt, so they can be tested against a remote DynamoDB. DynamoDBLocal chokes when provisionedThroughput is not set; provisioned throughput cannot be set if billing mode is set to PAY_PER_REQUEST, thus we can only test it against remote.

degaydorus:
@msgroi Updated ByTable to also allow for the billingMode to be set, even though it's a pass through, there's also an option to set it at the client level.  Let me know if this makes sense for ByTable. The tests I added for ByTable timeout on the clone step of createTable and I'm not sure why: https://github.com/salesforce/mt-dynamo/blob/78e22473c906481118492ad4f73b9ffdcbc12eff/src/main/java/com/salesforce/dynamodbv2/mt/mappers/MtAmazonDynamoDbByTable.java#L113.

src/main/java/com/salesforce/dynamodbv2/mt/mappers/sharedtable/SharedTableBuilder.java
```diff
@@ -149,12 +157,20 @@ protected void setDefaults() {
         if (this.defaultProvisionedThroughput == null) {
             this.defaultProvisionedThroughput = 1L;
         }
+        if (this.billingMode == null) {
+            this.billingMode = BillingMode.PROVISIONED;
+        }
         if (streamsEnabled == null) {
```
msgroi: Don't we want the default to be `PAY_PER_REQUEST`

degaydorus: I wasn't sure what default we wanted for mt-dynamo, since other clients like zos we will default to PAY_PER_REQUEST. I'll change the default in mt-dynamo to PAY_PER_REQUEST

msgroi: You're right.  I was thinking of the overall default.  At the mt-dynamo level, I suppose we should do whatever regular dynamo does.  That is, no default, it fails if you don't specify anything I think.

PR #273, opened 2019-02-06
extend EQ,GE support to EQ,GE,GT,LE,LT

horwitzj:


src/test/java/com/salesforce/dynamodbv2/QueryTest.java
```diff
@@ -142,8 +269,8 @@ private void queryWithKeyConditionsGtInner(TestArgument testArgument,
                         new Condition().withComparisonOperator(EQ).withAttributeValueList(createAttributeValue(
                             testArgument.getHashKeyAttrType(), HASH_KEY_VALUE)),
                         RANGE_KEY_FIELD,
-                        new Condition().withComparisonOperator(GT).withAttributeValueList(createAttributeValue(
-                            N, gtRValue))))
+                        new Condition().withComparisonOperator(op).withAttributeValueList(createAttributeValue(
+                            N, valueForComparisonOperator))))
                 ).getItems();
```
msgroi: All of the tests are actually testing varying the operator on range key fields.   Maybe the test names should indicate this?

horwitzj: in what way? there are currently 12 changed/created tests here: `queryWithKeyConditions[Gt|Ge|Lt|Le][Low|Middle|High]`—admittedly it's two characters (of a heck of a lot) hidden inside the method name, but did you mean something aside from the `[Gt|Ge|Lt|Le]` part of those method names? (note that `queryWithKeyConditionsComparisonOperatorInner` is the inner method that all 12 use at their core.)

PR #289, opened 2019-02-21
add query-paging support

horwitzj:


src/test/java/com/salesforce/dynamodbv2/QueryTest.java
```diff
@@ -100,178 +106,134 @@ void queryWithKeyConditionsEq(TestArgument testArgument) {
 
     // test legacy calls (see {@code QueryMapper} for more on "legacy".
     /**
-     * Table has (hk, rkLow), (hk, rkMiddle), (hk, rkHigh); we ask for items that match hk and have rk > rkLow, so there
```
horwitzj: instead of filling the table with three objects whose range keys are -1, 0, 1, it's an adjustable min and max (currently set at -1 and 10, respectively)—this was to allow for a paging test that uses a page size that isn't super small (i.e., 2 or 1)

src/test/java/com/salesforce/dynamodbv2/QueryTest.java
```diff
@@ -100,178 +106,134 @@ void queryWithKeyConditionsEq(TestArgument testArgument) {
...
+            IntStream.rangeClosed(RANGE_KEY_N_MIN + 2, RANGE_KEY_N_MAX)
+                .mapToObj(String::valueOf)
+                .collect(Collectors.toSet()));
     }
 
     /**
-     * Table has (hk, rkLow), (hk, rkMiddle), (hk, rkHigh); we ask for items that match hk and have rk > rkHigh, so
-     * there should be 0 results.
-     */
-    @ParameterizedTest(name = "{arguments}")
```
horwitzj: used to be 3 tests per operand (comparing with -1, comparing with 0, and comparing with -1—the three possible range keys), instead we now just have two per operand: compare with `RANGE_KEY_N_MIN` and compare with `RANGE_KEY_N_MIN + 1`

src/test/java/com/salesforce/dynamodbv2/QueryTest.java
```diff
@@ -100,178 +106,134 @@ void queryWithKeyConditionsEq(TestArgument testArgument) {
...
-                        HASH_KEY_FIELD,
-                        new Condition().withComparisonOperator(EQ).withAttributeValueList(createAttributeValue(
-                            testArgument.getHashKeyAttrType(), HASH_KEY_VALUE)),
-                        RANGE_KEY_FIELD,
-                        new Condition().withComparisonOperator(op).withAttributeValueList(createAttributeValue(
-                            N, valueForComparisonOperator))))
-                ).getItems();
+            final QueryRequest queryRequest = new QueryRequest()
+                .withTableName(TABLE4)
+                .withKeyConditions(getKeyConditions(op, valueForComparisonOperator, testArgument.getHashKeyAttrType()));
```
horwitzj: pulled out a new method (`getKeyConditions`) b/c it's used in a new test too

src/test/java/com/salesforce/dynamodbv2/QueryTest.java
```diff
@@ -437,4 +411,52 @@ void queryLsi(TestArgument testArgument) {
                     .build()));
         });
     }
+
```
horwitzj: here's the (parameterized) test for the new code...

src/main/java/com/salesforce/dynamodbv2/mt/mappers/sharedtable/impl/MtAmazonDynamoDbBySharedTable.java
```diff
@@ -313,18 +313,32 @@ public PutItemResult putItem(PutItemRequest putItemRequest) {
...
+        // map table name
+        final QueryRequest clonedQueryRequest = queryRequest.clone();
+        clonedQueryRequest.withTableName(tableMapping.getPhysicalTable().getTableName());
 
         // map query request
-        tableMapping.getQueryAndScanMapper().apply(queryRequest);
+        //clonedQueryRequest.setExpressionAttributeNames(Optional.ofNullable(clonedQueryRequest.getFilterExpression())
+        //    .map(s -> new HashMap<>(clonedQueryRequest.getExpressionAttributeNames())).orElseGet(HashMap::new));
+        //clonedQueryRequest.setExpressionAttributeValues(Optional.ofNullable(clonedQueryRequest.getFilterExpression())
+        //    .map(s -> new HashMap<>(clonedQueryRequest.getExpressionAttributeValues())).orElseGet(HashMap::new));
```
busjaeger: Did you mean to leave this comments in here?

src/main/java/com/salesforce/dynamodbv2/mt/mappers/sharedtable/impl/MtAmazonDynamoDbBySharedTable.java
```diff
@@ -313,18 +313,32 @@ public PutItemResult putItem(PutItemRequest putItemRequest) {
...
+        tableMapping.getQueryAndScanMapper().apply(clonedQueryRequest);
+
+        // keep moving forward pages until we find at least one record for current tenant or reach end
```
busjaeger: We shouldn't have to do that for queries (and I am trying to get rid of it for scans, because unbounded looping in the client thread causes issues): queries specify a hash key and all results for a given hash key are for the same tenant. I think the previous version of the code is sufficient.

src/main/java/com/salesforce/dynamodbv2/mt/mappers/sharedtable/impl/MtAmazonDynamoDbBySharedTable.java
```diff
@@ -313,18 +313,32 @@ public PutItemResult putItem(PutItemRequest putItemRequest) {
...
+            && queryResult.getLastEvaluatedKey() != null) {
+            clonedQueryRequest.setExclusiveStartKey(queryResult.getLastEvaluatedKey());
+        }
 
         // map result
-        QueryResult queryResult = getAmazonDynamoDb().query(queryRequest);
-        queryResult.setItems(
-            queryResult.getItems().stream().map(item -> tableMapping.getItemMapper().reverse(item)).collect(toList()));
+        queryResult.setItems(queryResult.getItems().stream().map(tableMapping.getItemMapper()::reverse)
+            .collect(toList()));
```
busjaeger: I like using method references to avoid lambdas :)

src/test/java/com/salesforce/dynamodbv2/testsupport/DefaultTestSetup.java
```diff
@@ -116,24 +116,12 @@ public void setupTableData(AmazonDynamoDB amazonDynamoDb,
...
-                                .rangeKey(N, RANGE_KEY_MIDDLE_N_VALUE)
-                                .build()));
-                    amazonDynamoDb.putItem(
-                        new PutItemRequest().withTableName(table)
-                            .withItem(ItemBuilder.builder(hashKeyAttrType, HASH_KEY_VALUE)
-                                .someField(S, SOME_OTHER_FIELD_VALUE + table + org)
-                                .rangeKey(N, RANGE_KEY_HIGH_N_VALUE)
-                                .build()));
+                                .rangeKey(N, String.valueOf(i))
+                                .build())));
```
busjaeger: Is this going to make our test cases a lot slower? (I generally feel torn about having the test parameter generation also initialize the fixtures.)

src/main/java/com/salesforce/dynamodbv2/mt/mappers/sharedtable/impl/MtAmazonDynamoDbBySharedTable.java
horwitzj: for a bit, I was trying to make scan and query code similarish, so I could extract common method(s)..

src/test/java/com/salesforce/dynamodbv2/testsupport/DefaultTestSetup.java
horwitzj: are you worried about overhead like creating the stream, converting `int`s to `String`s, etc., or just that `RANGE_KEY_N_MAX - RANGE_KEY_N_MIN + 1` could be big? (here it's 10-(-1)+1 = 12—so 12 entries are added (instead of the 3 that occur in master).)

could we lazily add these elements? (I mean, tests that don't access TABLE4 don't need it populated.)

busjaeger: Just that adding anything to this initialization logic multiplies by test and parameter combination. So small changes quickly balloon. Another option would be to fill the tables with more records during query test setup. If it's not too much overhead I am fine leaving it as-is; I was just curious how much slower this makes end-to-end mt-dynamo tests.

PR #315, opened 2019-03-20
beginnings of reproducing bug

msgroi:


src/test/java/com/salesforce/dynamodbv2/mt/util/CachingAmazonDynamoDbStreamsTest.java
```diff
@@ -820,6 +820,36 @@ void testDisableCache() {
...
+                .build();
+
+        String nextShardIterator = assertGetRecords(cachingStreams, firstRequest, null, 0, 6);
+
+        // Mock nextShardIterator result from first request and compare mock to result from previous assertGetRecords
+        GetShardIteratorRequest secondRequest = newAfterSequenceNumberRequest(5);
+        String secondIterator = mockGetShardIterator(streams, secondRequest);
+        mockGetRecords(streams, secondIterator, 5, 9);
+
+        assert (nextShardIterator.equals(cachingStreams.getShardIterator(secondRequest).getShardIterator()));
```
busjaeger: Should this be `assumeThat`?

degaydorus: We should keep it as assert for the following reason.  If it's `assumeThat`, then the old code would set nextShardIterator to the physical iterator and it wouldn't fail the test on assume rather than assert, which if the bug was ever accidentally introduced it could go undetected.

PR #353, opened 2019-04-17
Shared table getRecords improvements

busjaeger:
Need to write some more tests, but would be good to have you take a look to make sure the direction makes sense.

### DIFF
--- a/src/main/java/com/salesforce/dynamodbv2/mt/mappers/sharedtable/impl/MtAmazonDynamoDbStreamsBySharedTable.java
+++ b/src/main/java/com/salesforce/dynamodbv2/mt/mappers/sharedtable/impl/MtAmazonDynamoDbStreamsBySharedTable.java
@@ -65,8 +65,7 @@ public class MtAmazonDynamoDbStreamsBySharedTable extends MtAmazonDynamoDbStream
 
         final MtGetRecordsResult result = new MtGetRecordsResult()
             .withRecords(new ArrayList<>(limit))
-            .withNextShardIterator(request.getShardIterator())
-            .withLastSequenceNumber(null);
+            .withNextShardIterator(request.getShardIterator());
 
         while (addRecords(result, limit, recordMapper, recordFilter)
             && result.getRecords().size() < limit
--- a/src/main/java/com/salesforce/dynamodbv2/mt/mappers/sharedtable/impl/MtAmazonDynamoDbStreamsBySharedTable.java
+++ b/src/main/java/com/salesforce/dynamodbv2/mt/mappers/sharedtable/impl/MtAmazonDynamoDbStreamsBySharedTable.java
@@ -68,12 +68,10 @@ public class MtAmazonDynamoDbStreamsBySharedTable extends MtAmazonDynamoDbStream
             .withNextShardIterator(request.getShardIterator())
             .withLastSequenceNumber(null);
 
-        while (result.getRecords().size() < limit
+        while (addRecords(result, limit, recordMapper, recordFilter)
+            && result.getRecords().size() < limit
             && result.getNextShardIterator() != null
-            && (mtDynamoDb.getClock().millis() - time) < timeLimit) {
-            if (!addRecords(result, limit, recordMapper, recordFilter)) {
-                break;
-            }
+            && (mtDynamoDb.getClock().millis() - time) <= timeLimit) {
         }
 
         return result;
--- a/src/main/java/com/salesforce/dynamodbv2/mt/mappers/sharedtable/impl/MtAmazonDynamoDbStreamsBySharedTable.java
+++ b/src/main/java/com/salesforce/dynamodbv2/mt/mappers/sharedtable/impl/MtAmazonDynamoDbStreamsBySharedTable.java
@@ -1,7 +1,8 @@
 package com.salesforce.dynamodbv2.mt.mappers.sharedtable.impl;
 
 import static com.google.common.base.Preconditions.checkArgument;
-import static java.util.stream.Collectors.toCollection;
+import static com.google.common.base.Preconditions.checkState;
+import static com.google.common.collect.Iterables.getLast;
 
 import com.amazonaws.services.dynamodbv2.AmazonDynamoDBStreams;
 import com.amazonaws.services.dynamodbv2.model.AttributeValue;
@@ -31,10 +32,10 @@ public class MtAmazonDynamoDbStreamsBySharedTable extends MtAmazonDynamoDbStream
      * Default constructor.
      *
      * @param dynamoDbStreams underlying streams instance
-     * @param mtDynamoDb corresponding shared table dynamo DB instance
+     * @param mtDynamoDb      corresponding shared table dynamo DB instance
      */
     public MtAmazonDynamoDbStreamsBySharedTable(AmazonDynamoDBStreams dynamoDbStreams,
-        MtAmazonDynamoDbBySharedTable mtDynamoDb) {
+                                                MtAmazonDynamoDbBySharedTable mtDynamoDb) {
         super(dynamoDbStreams, mtDynamoDb);
     }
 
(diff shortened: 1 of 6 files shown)


src/main/java/com/salesforce/dynamodbv2/mt/mappers/sharedtable/impl/MtAmazonDynamoDbStreamsBySharedTable.java
```diff
@@ -43,30 +44,108 @@ public GetShardIteratorResult getShardIterator(GetShardIteratorRequest request)
...
-            List<Record> records = result.getRecords();
-            if (records.isEmpty()) {
-                break;
-            }
-            records.stream().map(recordMapper).filter(recordFilter).collect(toCollection(() -> mtRecords));
-        } while (mtRecords.isEmpty() && iterator != null);
+        final MtGetRecordsResult result = new MtGetRecordsResult()
+            .withRecords(new ArrayList<>(limit))
+            .withNextShardIterator(request.getShardIterator())
+            .withLastSequenceNumber(null);
```
degaydorus: lastSequenceNumber by default will be null, is this line needed?
`.withLastSequenceNumber(null);`

busjaeger: That's a good point. It's a remnant of a previous iteration where I had explicit variables for each and was creating the result after the loop. I'll take this line out.

src/main/java/com/salesforce/dynamodbv2/mt/mappers/sharedtable/impl/MtAmazonDynamoDbStreamsBySharedTable.java
```diff
@@ -43,30 +44,108 @@ public GetShardIteratorResult getShardIterator(GetShardIteratorRequest request)
...
+                               Function<Record, MtRecord> recordMapper,
+                               Predicate<MtRecord> recordFilter) {
+        // retrieve max number of records from underlying stream (and retry below if we got too many)
+        final GetRecordsRequest innerRequest = new GetRecordsRequest().withLimit(MAX_LIMIT)
+            .withShardIterator(result.getNextShardIterator());
+        GetRecordsResult innerResult = dynamoDbStreams.getRecords(innerRequest);
+        result.setNextShardIterator(innerResult.getNextShardIterator());
+        List<Record> records = innerResult.getRecords();
+        // if we did not any records, no point continuing
+        if (records.isEmpty()) {
```
degaydorus: there could be a NPE if `MtGetRecordsResult().withRecords()` is never called (based on records implementation in aws GetRecordsResult).  This is not the case for `getRecords` but just thinking ahead if MtGetRecordsResult is reused.

Might be helpful to add a default records value in MtGetRecordsResult, before doing operations on records?

src/main/java/com/salesforce/dynamodbv2/mt/mappers/sharedtable/impl/MtAmazonDynamoDbStreamsBySharedTable.java
```diff
@@ -43,30 +44,108 @@ public GetShardIteratorResult getShardIterator(GetShardIteratorRequest request)
...
+        final List<Record> innerMtRecords = new ArrayList<>(innerLimit);
+        int consumedRecords = addRecords(innerMtRecords, innerLimit, records, recordMapper, recordFilter);
+
+        // If we consumed all records, then we did not exceed the limit, so worth continuing (if we have time)
+        if (consumedRecords == records.size()) {
+            result.getRecords().addAll(innerMtRecords);
+            result.setLastSequenceNumber(getLast(records).getDynamodb().getSequenceNumber());
+            return true;
+        } else {
+            // If we exceeded limit: retry with lower limit (can't just return a subset of records, because the next
```
degaydorus: the else block will be reached if consumedRecords < records.size and consumedRecords > records.size

what is expected for this case? `consumedRecords < records.size`, the commend address this case I think `consumedRecords > records.size`

src/main/java/com/salesforce/dynamodbv2/mt/mappers/sharedtable/impl/MtAmazonDynamoDbStreamsBySharedTable.java
busjaeger: I am not quite sure I follow. This line calls a method on the result of calling `innerResult.getRecords()` which is returned by the underlying `dynamoDbStreams` implementation. How is this related to `MtGetRecordsResult`?

src/main/java/com/salesforce/dynamodbv2/mt/mappers/sharedtable/impl/MtAmazonDynamoDbStreamsBySharedTable.java
busjaeger: An invariant of `addRecords` is that `consumedRecords <= records.size()`, since `consumedRecords` is incremented at most once for each element of `records`. I suppose I could add an assert or something?

degaydorus: okay got it

src/main/java/com/salesforce/dynamodbv2/mt/mappers/sharedtable/impl/MtAmazonDynamoDbStreamsBySharedTable.java
degaydorus: Looks like I confused the two in the comment (records and input `MtGetRecordsResult` result). I was observing that if `MtGetRecordsResult: withRecords` is called and the input parameter has a value of `null`, then `MtGetRecordsResult: withRecords` calls aws `GetRecordsResult:setRecords` implementation and in this case it will set the `records` to `null`.  It's only if `GetRecordsResult:withRecords` is called, that a non null default is set.  It doesn't affect the implementation in `addRecords`, it is just something to keep in mind , which is not to pass in a null collection into `MtGetRecordsResult:withRecords`, since we're overriding the withRecords method, and calling `GetRecordsResult:setRecords` directly.  Was suggesting we handle null input param cases in `MtGetRecordsResult:withRecords`.

degaydorus: GetRecordsResult:setRecords:   
 public void setRecords(Collection<Record> records) {
        if (records == null) {
            **this.records = null;**
        } else {
            this.records = new ArrayList(records);
        }
    }

busjaeger: `GetRecordsResult` only defaults to an empty collection for `withRecords(Record... records)`, not for `withRecords(java.util.Collection<Record> records)`, correct?

degaydorus: then in `MtAmazonDynamoDbStreamsBySharedTable:addRecords(MtGetRecordsResult result,..)` 
`final int innerLimit = limit - result.getRecords().size();` , if records are null, result.getRecords() would be null

busjaeger: `result#records` is initialized to an empty list on line 67. How could `result.getRecords()` be null? Or are you just talking purely from the perspective of this method?

I think in general, I am assuming an invariant that records are never null, because either our code or the sdk initializes it to a (possibly empty) collection. You make a good point though, that maybe we should be programming more defensively?

degaydorus: yes `result#records` is initialized to an empty list on line 67 , so it's handled for this merge.  Mainly commenting on `MtGetRecordsResult: withRecords` usage, and I noticed this case when reviewing  `MtAmazonDynamoDbStreamsBySharedTable:addRecords(MtGetRecordsResult result,..)`, so we're good for this, but since `MtGetRecordsResult: withRecords` is public, a user of mt-dynamo who calls `MtGetRecordsResult: withRecords` could run into this if input is `null`.

Yes, from what I saw in aws's implementation, they don't default initialize unless `GetRecordsResult withRecords(Record... records)` which is the concern since we are overriding the other `withRecords` method that sets null to null.

busjaeger: The request and result objects in the AWS SDK are just DTOs. There are no cases where clients build a result object and pass it into the dynamo sdk; it's always the sdk building a result object and returning it to clients. You could argue that they shouldn't expose setters and maybe even make them immutable objects, but presumably they had their reasons for this approach. It does make things like unit testing (mocking) easier.

mt-dynamo follows the same pattern: it exposes a mutable `MtGetRecordsResult` object, but only exposes methods that return instances of it (none that take it as an input argument).

PR #375, opened 2019-05-09
integrate kotlin into the build - baby steps

pallavkothari:
all I'm doing here is integrating kotlin into the build - leaving the door open for endless possibilities 🙏


sbabu-salesforce:
rather than muddy the code with commented out pom.xml snippets, I filed an issue to this repo: https://github.com/salesforce/mt-dynamo/issues/462

@horwitzj  hope that explains why I pulled support for generating html docs. Had side conversations with Mike and PK, and agreed its better to pull javadoc generation support, versus wrestling with dokka to get parity with before.

pom.xml
```diff
@@ -23,6 +23,10 @@
         <junit.version>5.4.2</junit.version>
         <aws.sdk.version>1.11.591</aws.sdk.version>
         <skipITs>true</skipITs>
+        <kotlin.version>1.3.31</kotlin.version>
```
horwitzj: any reason not to use 1.3.41? (and spotless 1.23.1, ktlint 0.33.0)

sbabu-salesforce: no reason... does dependabot not handle that?

horwitzj: indeed it might, though dependabot doesn't catch everything (and I happened to notice those three)

PR #411, opened 2019-06-06
Mt scan support

sbabu-salesforce:


src/main/java/com/salesforce/dynamodbv2/mt/mappers/MtAmazonDynamoDb.java
```diff
@@ -95,4 +97,33 @@ public String toString() {
         }
     }
 
+    /**
+     * Wrap a scan result from a multi tenant scan, to associate tenant information per scan result item.
+     **/
+    class MtScanResult extends ScanResult {
```
sbabu-salesforce: not crazy about how this is done, which is fragile if ScanResult gets changed in a newer aws version, open to feedback

src/main/java/com/salesforce/dynamodbv2/mt/mappers/MtAmazonDynamoDbBase.java
```diff
@@ -129,7 +129,7 @@ protected MtAmazonDynamoDbContextProvider getMtContext() {
      * @param tableName Name of the table.
      * @return true if the given table name is a multitenant table associated with this instance, false otherwise.
      */
-    protected boolean isMtTable(String tableName) {
+    public boolean isMtTable(String tableName) {
```
sbabu-salesforce: needed this sort of call from the test.... I explore trying to delete all tables between test runs, so a listTables() command doesn't return tables from other test runs, but didn't have much luck. Gave up trying to clean up other tests' tables, and tried filtering to just the tables this mt-dynamo client manages instead

src/main/java/com/salesforce/dynamodbv2/mt/mappers/MtAmazonDynamoDbByTable.java
```diff
@@ -215,9 +217,22 @@ public QueryResult query(QueryRequest queryRequest) {
...
     public ScanResult scan(ScanRequest scanRequest) {
-        scanRequest = scanRequest.clone();
-        scanRequest.withTableName(buildPrefixedTableName(scanRequest.getTableName()));
-        return getAmazonDynamoDb().scan(scanRequest);
+        if (getMtContext().getContextOpt().isEmpty()) {
+            Preconditions.checkArgument(isMtTable(scanRequest.getTableName()));
+            ScanResult result =  getAmazonDynamoDb().scan(scanRequest);
+            String[] tenantTable = getTenantAndTableName(scanRequest.getTableName());
+            List<String> tenants =new ArrayList(result.getCount());
+            List<String> tables = new ArrayList(result.getCount());
```
sbabu-salesforce: somewhat silly to build a huge list of the same items, but I figure, its better to stay compatible across byTable or bySharedTable with the same MtScanResult object def

src/test/java/com/salesforce/dynamodbv2/testsupport/ArgumentBuilder.java
```diff
@@ -66,13 +67,13 @@ public ArgumentBuilder(AmazonDynamoDB rootAmazonDynamoDb) {
...
     @Override
     public List<TestArgument> get() {
-        return getAmazonDynamoDbStrategies().stream().flatMap(
-            (Function<AmazonDynamoDB, Stream<TestArgument>>) amazonDynamoDB ->
-                getHashKeyAttrTypes().stream().flatMap(
-                    (Function<ScalarAttributeType, Stream<TestArgument>>) scalarAttributeType ->
-                        Stream.of(new TestArgument(amazonDynamoDB,
-                            getOrgs(),
-                            scalarAttributeType)))).collect(Collectors.toList());
+        List<TestArgument> ret = Lists.newArrayList();
```
sbabu-salesforce: no functional change here, was tinkering with this to add managed tables to TestArgument. Backtracked that approach, to instead make isMtTable() public, but left this refactoring in, just since I find the flatMap stream approach much less readable given what the streams are doing.

msgroi: :thumbsup:

src/main/java/com/salesforce/dynamodbv2/mt/mappers/sharedtable/impl/MtAmazonDynamoDbBySharedTable.java
```diff
@@ -390,10 +392,44 @@ public QueryResult query(QueryRequest queryRequest) {
...
+            List<String> tenants = new ArrayList<>(scanResult.getItems().size());
+            List<String> virtualTables = new ArrayList<>(scanResult.getItems().size());
+            ScanResult ret = new MtScanResult(scanResult, tenants, virtualTables);
+            for (Map<String, AttributeValue> item : scanResult.getItems()) {
+                // go through each row in the scan, and pull out the tenant and table information from the primary key
+                // to separate attributes in each items map
+                FieldValue virtualFieldKeys = fieldMapperFunction.apply(item);
+                TableMapping tableMappping = getMtContext().withContext(virtualFieldKeys.getContext(),
+                    tableName -> getTableMapping(tableName), virtualFieldKeys.getTableName());
+                Map<String, AttributeValue> unpackedVirtualItem = tableMappping.getItemMapper().reverse(item);
```
msgroi: Are you handling paging(`lastEvaluatedKey`)?

msgroi: Is there enough commonality with the loop at line 462 below to have some common code?

sbabu-salesforce: so I wanna say that's being delegated to default scan() behavior, so I don't expect that will break. This scan is taking the exact same scan request and result from the lower scanRequest call, and just transforming the item map to use the same keys the tenant created the virtual table metadata with, along side tenant table information in the MtScanResult. Will run into that soon if it does break, when actually trying to use paged scanning to backup a large table.

sbabu-salesforce: I gave it a try to pull out duplicated logic into a method, and didn't get more than one line out to the shared method, so gave up.

src/main/java/com/salesforce/dynamodbv2/mt/mappers/MtAmazonDynamoDb.java
busjaeger: Ideally, we'd want tenant and table name attached to each time in the scan result, right? Have you considered putting items with special keys into the map, such as `mt:contxt` and `mt:tableName`?

src/main/java/com/salesforce/dynamodbv2/mt/mappers/MtAmazonDynamoDbBase.java
```diff
@@ -329,7 +329,7 @@ public ListTablesResult listTables() {
 
     @Override
     public ListTablesResult listTables(String exclusiveStartTableName) {
-        throw new UnsupportedOperationException();
+        return getAmazonDynamoDb().listTables(exclusiveStartTableName);
```
busjaeger: Shouldn't this only return tables visible to the current context? I.e., if you are calling with context "tenant1", only "tenant1" tables. And if you are calling with no context (your use case) only tables for which `isMtTable` returns true?

src/main/java/com/salesforce/dynamodbv2/mt/mappers/sharedtable/impl/MtAmazonDynamoDbBySharedTable.java
```diff
@@ -401,10 +403,17 @@ public QueryResult query(QueryRequest queryRequest) {
     }
 
     /**
-     * TODO: write Javadoc.
+     * Execute a scan meeting the specs of @{link AmazonDynamoDB}, but scoped to single tenants within a shared table.
```
horwitzj: "@{link" -> "{@link"

src/main/java/com/salesforce/dynamodbv2/mt/mappers/sharedtable/impl/MtAmazonDynamoDbBySharedTable.java
```diff
@@ -401,10 +403,17 @@ public QueryResult query(QueryRequest queryRequest) {
...
+     * If no tenant context is specified, a multi tenant scan is performed over the multi tenant shared table.
```
horwitzj: "multi tenant" -> "multitenant" (here and anywhere else; sim. with "multi-tenant" -> "multitenant")

src/main/java/com/salesforce/dynamodbv2/mt/mappers/sharedtable/impl/MtAmazonDynamoDbBySharedTable.java
```diff
@@ -401,10 +403,17 @@ public QueryResult query(QueryRequest queryRequest) {
...
+     * Scans scoped to single tenants is not a performant operation.
```
horwitzj: "is not a performant operation." -> "are not performant operations." (or, better (IMO): "are inefficient.")

src/main/java/com/salesforce/dynamodbv2/mt/mappers/sharedtable/impl/MtAmazonDynamoDbBySharedTable.java
```diff
@@ -401,10 +403,17 @@ public QueryResult query(QueryRequest queryRequest) {
...
+     * Multi tenant scans are as bad as scans on any other dynamo table, ie: not great.
```
horwitzj: "ie:" -> "i.e.,"

src/test/java/com/salesforce/dynamodbv2/ScanTest.java
```diff
@@ -135,6 +138,45 @@ void scanAll(TestArgument testArgument) {
...
+    void scanAllTenants(TestArgument testArgument) {
+        MT_CONTEXT.setContext(null);
+        String startTableName = null;
+        List<String> tableNames;
+        int maxPages = 5;
+        int attemptCount = 0;
+        do {
+            List<String> allTableNames = testArgument.getAmazonDynamoDb().listTables(startTableName).getTableNames();
+
+            //filter scan to just tables this mt dymamo strategy manages
```
horwitzj: -> "// filter scan to just tables managed by this mt-dynamo strategy"

src/test/java/com/salesforce/dynamodbv2/ScanTest.java
```diff
@@ -135,6 +138,45 @@ void scanAll(TestArgument testArgument) {
...
+            tableNames = allTableNames.stream()
+                .filter(t -> ((MtAmazonDynamoDbBase) testArgument.getAmazonDynamoDb()).isMtTable(t))
+                .collect(Collectors.toList());
+            startTableName = allTableNames.get(allTableNames.size() - 1);
+            attemptCount++;
+        } while (tableNames.size() == 0 && attemptCount < maxPages);
+
+        assertTrue(tableNames.size() > 0, "No managed tables found to scan, strange..");
+        // go through every table, and issue at least one successful scan request.
+        // validate at least one table is populated
```
horwitzj: add "that": "// validate that"

src/test/java/com/salesforce/dynamodbv2/ScanTest.java
horwitzj: (FYI: Intellij's spell checker should catch "dymamo" (though it probably won't complain about "mt dynamo" (as opposed to "my-dynamo"))—and other potential issues)

src/main/java/com/salesforce/dynamodbv2/mt/mappers/MtAmazonDynamoDbBase.java
busjaeger: Are you trying to delete virtual or physical tables in your test?

sbabu-salesforce: its physical tables left behind from other test cases that's the problem. when using raw `amazonDb.listTables()` to try running a scan against all tables, I'm getting plenty of physical tables leaked from other tests. I made this public, so I can filter all the physical tables from raw `AmazonDb.listTables()` to the set of mtTables that each given implementation of mt-dynamo (byTable, byIndex, byAccount) actually is using

busjaeger: makes sense. `MtAmazonDynamoDbStreamsBaseTest#deleteMtTables` is doing that too, but it was probably already able to access that method, because it's in the same package.

busjaeger: We could also consider implementing `listTables` with and without context, similar to what we're doing for scans and `listStreams` (though the latter currently only supports without context).

src/main/java/com/salesforce/dynamodbv2/mt/mappers/MtAmazonDynamoDbBase.java
```diff
@@ -324,12 +330,20 @@ public ListTablesResult listTables(ListTablesRequest listTablesRequest) {
 
     @Override
     public ListTablesResult listTables() {
-        return getAmazonDynamoDb().listTables();
+        if (mtContext.getContextOpt().isEmpty()) {
+            return getAmazonDynamoDb().listTables();
```
busjaeger: I think this method would need to be implemented in the subclasses to take prefixes and statically configured tables into account.

src/main/java/com/salesforce/dynamodbv2/mt/mappers/sharedtable/impl/MtAmazonDynamoDbBySharedTable.java
```diff
@@ -444,6 +453,31 @@ public ScanResult scan(ScanRequest scanRequest) {
         return scanResult;
     }
 
+    private ScanResult multiTenantScan(ScanRequest scanRequest) {
```
busjaeger: Maybe `crossTenantScan` or `scanAll` (I also struggled to come up with a good name; see https://github.com/salesforce/mt-dynamo/blob/master/src/main/java/com/salesforce/dynamodbv2/mt/mappers/MtAmazonDynamoDbStreamsBase.java#L202).

src/test/java/com/salesforce/dynamodbv2/ScanTest.java
```diff
@@ -135,6 +138,45 @@ void scanAll(TestArgument testArgument) {
...
+        String startTableName = null;
+        List<String> tableNames;
+        int maxPages = 5;
+        int attemptCount = 0;
+        do {
+            List<String> allTableNames = testArgument.getAmazonDynamoDb().listTables(startTableName).getTableNames();
+
+            //filter scan to just tables this mt dymamo strategy manages
+            tableNames = allTableNames.stream()
+                .filter(t -> ((MtAmazonDynamoDbBase) testArgument.getAmazonDynamoDb()).isMtTable(t))
```
busjaeger: Seems like `listTables` should do this filtering internally (and then maybe we don't not need to make `isMtTable` public).

src/main/java/com/salesforce/dynamodbv2/mt/mappers/sharedtable/impl/MtAmazonDynamoDbBySharedTable.java
sbabu-salesforce: "ie:" is shorter and gets the same message across. Agree to disagree?

src/main/java/com/salesforce/dynamodbv2/mt/mappers/sharedtable/impl/MtAmazonDynamoDbBySharedTable.java
sbabu-salesforce: How about `scanAllTenants`?

src/main/java/com/salesforce/dynamodbv2/mt/mappers/MtAmazonDynamoDbBase.java
sbabu-salesforce: actually, by filtering with `isMtTable` here, that should be handled for us? That solves both 👍

src/main/java/com/salesforce/dynamodbv2/mt/mappers/sharedtable/impl/MtAmazonDynamoDbBySharedTable.java
horwitzj: the Salesforce Style Guide (as well basically every other style guide (including the broadly accepted Chicago Manual of Style)) insists on "i.e." as the spelling for the abbreviation of "id est" (and also insists that "i.e." should be followed by a comma). (FWIW, the Salesforce Style Guide goes one further and at least suggests avoiding "i.e." entirely—e.g., replacing it with something like "that is" or "in other words".)

while the actual code is where the most important work is, documentation—especially public-facing (open-source) documentation—adhering to a uniform style enhances clarity and readability.

sbabu-salesforce:
@busjaeger  can you take another pass through this PR and merge it if it looks okay?

src/main/java/com/salesforce/dynamodbv2/mt/mappers/MtAmazonDynamoDbBase.java
```diff
@@ -332,22 +340,40 @@ public ListTablesResult listTables(ListTablesRequest listTablesRequest) {
...
+            // and eagerly pull more pages until we fill up a full result set or run out of tables
+            do  {
+                ListTablesResult rawResults = getAmazonDynamoDb().listTables(innerExclusiveStartTableName, limit);
+                tableNames.addAll(
+                    rawResults.getTableNames().stream().filter(this::isMtTable).collect(Collectors.toList()));
+                innerExclusiveStartTableName = rawResults.getLastEvaluatedTableName();
+            } while (!(tableNames.size() >= limit || innerExclusiveStartTableName == null));
+
+            return new ListTablesResult()
+                .withLastEvaluatedTableName(tableNames.isEmpty() ? null : tableNames.get(tableNames.size() - 1))
```
busjaeger: Shouldn't this always be null? (Since you are going through all tables at once.)

src/main/java/com/salesforce/dynamodbv2/mt/mappers/MtAmazonDynamoDbBase.java
```diff
@@ -107,6 +109,12 @@
  */
 public class MtAmazonDynamoDbBase implements MtAmazonDynamoDb {
 
+    /**
+     * Special "column" key returned to client on multitenant scans.
+     */
+    public static final String TENANT_KEY = "mt:context";
+    public static final String VIRTUAL_TABLE_KEY = "mt:tableName";
```
busjaeger: It would probably be good to make these attribute names configurable, so that clients can assign strings that they know for sure won't conflict with other physical or virtual attribute names.

busjaeger:
@sbabu-salesforce we've been following the workflow of having the author merge changes once approved (in ZOS as well), so feel free to merge when it's convenient for you. (We can also discuss changing that workflow if it's uncommon.)

src/main/java/com/salesforce/dynamodbv2/mt/mappers/MtAmazonDynamoDbBase.java
sbabu-salesforce: it'll be non-null if tableNames fills its result set (limit tables scanned), which this test case exercises: https://github.com/salesforce/mt-dynamo/pull/411/files/c1163ffff0dfaea05d8c86e6a0bfed921e115c69#diff-ecb914953bc1a215b4e5c00fdeb28cafR30

src/main/java/com/salesforce/dynamodbv2/mt/mappers/MtAmazonDynamoDbBase.java
sbabu-salesforce: sounds good, will include that in a followup OR

src/main/java/com/salesforce/dynamodbv2/mt/mappers/MtAmazonDynamoDbBase.java
busjaeger: Then shouldn’t the condition be for whether to set last evaluated table name the same as in the next line? (Ie set to non-null if the names exceed limit)

sbabu-salesforce: good catch, this doesn't match API contract of leaving lastEvaluatedTableName to null if there's no more tables to return. Will include that in a follow-up PR as well

sbabu-salesforce: addressed this in https://github.com/salesforce/mt-dynamo/pull/472

PR #424, opened 2019-06-19
Reduce get records calls

busjaeger:
I have been working a little here and there (mostly on flights) on a possible improvement to the stream caching for shared tables. It's not ready to be merged, because unit tests are still failing (expected) and some boundary conditions most likely aren't handled correctly. Nevertheless, it's in a state that folks can review and possibly contribute to (e.g., by adjusting test cases), so wanted to put a PR out there.

As discussed with @msgroi before I went on PTO, the basic idea is to have the caching layer decide when to query the underlying stream and only have the mt layer continue to query if the max number of records was returned. I also added a shard-level lock to avoid concurrent calls to the stream to avoid hitting limits and unnecessary calls. I also made some algorithmic improvements to reduce the amount of computation done in the caching layer.

It's hard to say how much these changes will help without having detailed metrics. It seems like y'all have been doing a bunch of work around that recently, so maybe we already have more visibility. Either way, it seems worth getting this PR into a stable state (i.e., fixing/adding test cases) and then giving it a try to compare against the current implementation.


src/main/java/com/salesforce/dynamodbv2/mt/util/CachingAmazonDynamoDbStreams.java
```diff
@@ -592,36 +414,30 @@ public int hashCode() {
         }
     }
 
-    // logger instance
-    private static final Logger LOG = LoggerFactory.getLogger(CachingAmazonDynamoDbStreams.class);
+    private static class CachingGetRecordsResult {
+        private final List<Record> records;
+        private final CachingShardIterator nextShardIterator;
```
shirdoo: doesn't need to be in this PR, but I'd also like to see stats of cache hits vs misses (how many stream callouts were made vs how much came off the cache), and given we have potential lock wait states, timing information for those wait states

src/main/java/com/salesforce/dynamodbv2/mt/util/RecordsCache.java
```diff
@@ -0,0 +1,276 @@
...
+
+    // config parameter
+    private final long maxRecordsByteSize;
+
+    // cached record segments sorted by sequence number within each shard
+    private final Map<ShardId, NavigableMap<SequenceNumber, Segment>> segments;
+    // Insertion order of cache segments for eviction purposes
+    private final LinkedList<ShardLocation> insertionOrder;
+    // looks for mutating lock
+    private final ReadWriteLock lock;
```
shirdoo: this is a JVM global lock for any mutations to cached segments? or is it more granular?

src/main/java/com/salesforce/dynamodbv2/mt/util/RecordsCache.java
```diff
@@ -0,0 +1,276 @@
...
+        } finally {
+            readLock.unlock();
+        }
+    }
+
+    // Should we bring back segment merging to avoid cache fragmentation?
+    void putRecords(ShardLocation shardLocation, List<Record> records) {
+        assert shardLocation != null && records != null && !records.isEmpty();
+
+        final Lock writeLock = lock.writeLock();
```
shirdoo: is there any potential of web requests waiting on this lock? If so, it should be timed, and returned as metadata back to the client (ZOS) in whatever Result object is being passed back. When we cross over SLA's, we need to be able to track down the bottlenecks, and I'm worried newrelic isn't granular enough to do that for us, with its transaction sampling

src/main/java/com/salesforce/dynamodbv2/mt/util/RecordsCache.java
shirdoo: Oh, I see, given we have a max recordsByteSize across this cache, we have to enforce a JVM global lock, so insertions into the cache properly evict the oldest segments inserted into the cache, across all tenants and shards. 

I'm worried this cache will get so much churn from typical stream processing and delta API requests, that the write lock will begin throttling readers of the cache. Another snippet worth instrumenting to return back to ZOS, as a potential bottleneck of delta API queries.

src/main/java/com/salesforce/dynamodbv2/mt/util/CachingAmazonDynamoDbStreams.java
```diff
@@ -724,14 +531,83 @@ public GetRecordsResult getRecords(GetRecordsRequest request) {
...
+        final CachingShardIterator nextIterator;
+        try {
+            // fetch records from cache
+            iterator.resolveLocation()
+                .map(location -> recordCache.getRecords(location, limit))
+                .ifPresent(records::addAll);
+
+            // check if we got enough
+            if (records.size() < limit) {
+                // not enough cached records: read segment from underlying shard
```
shirdoo: the number of delta API requests that step into this block needs to be bubbled up to to ZOS so we can understand the ratio of queries that are served off this cache exclusively, to the ratio of queries that hit the underlying stream, to be reloaded, and the cost to updating the cache. 

Given this cache is homegrown, we can't rely on the Guava cache monitoring we have to get things like hitCount, loadTime, evictionCount, etc.. so we have to instrument those pieces ourselves. Given how critical this cache is to a performant delta API, we need those metrics

shirdoo:
Most of my comments of needing more metrics bubbled up to ZOS are not directly related to these changes, mostly just from understanding the code being refactored. 

I think the functional changes of not querying more unless records are maxed out is the real potential silver bullet, and the rest of the refactorings look non functional for the most part looks good. We're getting a steady stream of 503's for delta api queries on perf (albeit a relatively low rate, ie ~5%), as well as slow performing queries, so we'll see if this change helps that. Looks good!

src/main/java/com/salesforce/dynamodbv2/mt/util/CachingAmazonDynamoDbStreams.java
busjaeger: One of the things we were discussing before I left was defining some kind of metrics interface for mt-dynamo that users of the library could supply. Is something like that in progress as part of the metrics work you guys have been doing?

src/main/java/com/salesforce/dynamodbv2/mt/util/CachingAmazonDynamoDbStreams.java
busjaeger: Agree. There are also some more lower-level stats at the caching layer we may want to emit, like how many cache segments we iterated over to collect the result.

This is one of the functional changes compared to the version currently in master: in master the caching adapter returns immediately if it finds any number of records in the cache (see https://github.com/salesforce/mt-dynamo/blob/master/src/main/java/com/salesforce/dynamodbv2/mt/util/CachingAmazonDynamoDbStreams.java#L758). And the multitenant adapter continues to call `getRecords` if it got any back (https://github.com/salesforce/mt-dynamo/blob/master/src/main/java/com/salesforce/dynamodbv2/mt/mappers/sharedtable/impl/MtAmazonDynamoDbStreamsBySharedTable.java#L106).

With this change, the caching layer tries to return as many records as possible, by iterating over adjacent cache segments (https://github.com/salesforce/mt-dynamo/pull/424/files#diff-144e0b563c0b1615fa0417b18acc0778R194) and then makes a single attempt to fetch more records from the underlying stream if it didn't get enough. The multitenant adapter only continues to call `getRecords` if it retrieved the maximum number of records (https://github.com/salesforce/mt-dynamo/pull/424/files#diff-a48916ee1793548f6147598a5d2fcf1eR106).

DynamoDB docs actually state "GetRecords can retrieve a maximum of 1 MB of data or 1000 stream records, whichever comes first." (https://docs.aws.amazon.com/amazondynamodb/latest/APIReference/API_streams_GetRecords.html), so we may want/need to consider checking for 1MB or 1k.

src/main/java/com/salesforce/dynamodbv2/mt/util/CachingAmazonDynamoDbStreams.java
```diff
@@ -801,66 +664,25 @@ private GetRecordsResult getRecords(CachingShardIterator iterator) {
...
-            GetRecordsResult result;
-            final Lock writeLock = recordsCacheLock.writeLock();
-            writeLock.lock();
-            try {
-                // Add retrieved records to cache under resolved position
-                cachedResult = addToCache(loadedPosition, loadedRecordsResult);
-
-                // now lookup result: may not be exactly what we loaded if we merged result with other segments.
-                // can also be empty if the cache is too small to hold the result, in which case we just return loaded.
-                result = getFromCache(loadedPosition).orElseGet(() ->
```
busjaeger: This behavior has changed as well: instead of adding to the cache and then doing another cache lookup, the new impl adds the retrieved records the records it found in the cache (without an extra lookup).

src/main/java/com/salesforce/dynamodbv2/mt/util/RecordsCache.java
```diff
@@ -0,0 +1,276 @@
...
+            final NavigableMap<SequenceNumber, Segment> shardCache = segments.get(shardId);
+            if (shardCache == null) {
+                previousSegment = null;
+                nextSegment = null;
+            } else {
+                previousSegment = getValue(shardCache.floorEntry(sequenceNumber));
+                nextSegment = getValue(shardCache.higherEntry(sequenceNumber));
+            }
+
+            final Segment segment = new Segment(sequenceNumber, records).subSegment(previousSegment, nextSegment);
```
busjaeger: This is another behavior change compared to master: this implementation does not merge segments (see https://docs.aws.amazon.com/amazondynamodb/latest/APIReference/API_streams_GetRecords.html). It seems less important to merge, given that the read code now iterates over adjacent segments. But we may still want to evaluate/measure the tradeoffs. For example, not merging likely makes writes faster, reducing the time in the critical section. On the other hand, having too many small segments could increase the time it takes to read and evict segments, since more lookups in the sorted map are needed.

src/main/java/com/salesforce/dynamodbv2/mt/util/RecordsCache.java
busjaeger: Great questions and observations! Exactly the discussion I was hoping to have with you guys.

In this new version, there are two locks: one is a shard-level lock (https://github.com/salesforce/mt-dynamo/pull/424/files#diff-90d28430b63aa6120b6b0ce714d31748R548) and one is a cache-level lock (this one).

The shard-level lock is not strictly necessary and we currently do not use one in master (see https://github.com/salesforce/mt-dynamo/blob/master/src/main/java/com/salesforce/dynamodbv2/mt/util/CachingAmazonDynamoDbStreams.java#L749). As stated in the comment (https://github.com/salesforce/mt-dynamo/pull/424/files#diff-90d28430b63aa6120b6b0ce714d31748R542), the purpose of the shard-level lock is to sequence access to each underlying shard to avoid hitting DynamoDB limits or throttling (https://docs.aws.amazon.com/amazondynamodb/latest/developerguide/Limits.html#limits-dynamodb-streams)  and to avoid having concurrent readers (KCL or delta API) load overlapping segments into the cache. There are probably some alternative solutions for addressing each of these concerns (and it may be possible/preferable to address them separately). For example, may it would be sufficient to use a more granular lock at the level of a shard sequence number (or maybe range?) to avoid unnecessary concurrent loads and then just rely on DynamoDB streams (like we do in master) to sequence/throttle access to the shard.

The cache lock guards (1) the non-thread safe collections used by the cache implementation and (2) the invariant that no two cache segments overlap. The invariant is enforced, but checking both the pervious and next cache segments when adding a new one and dropping any reducing the range of the added segment to avoid overlap (https://github.com/salesforce/mt-dynamo/pull/424/files#diff-144e0b563c0b1615fa0417b18acc0778R230). That's what allows the `get` method to simply lookup the next segment by using the end of the current segment (https://github.com/salesforce/mt-dynamo/pull/424/files#diff-144e0b563c0b1615fa0417b18acc0778R195). Again, there are likely different ways to go about implementing this range cache. One of the main motivations for splitting this logic into a separate class is to make it easier for us to change it independently of the caching streams adapter. For example, one option we could explore is to relax the invariant of having no overlap. And then maybe we could use a standard Guava cache with a sorted index we maintain via an eviction lister to reduce the write contention. The assumption underlying the current implementation is that the time to update the cache data structure (i.e., the `putRecords` method) is orders of magnitude faster than loading records from the DynamoDB stream. And since we need to sequence stream reads to some extent anyway to avoid hitting limits, write contention on the in-memory data structure may not be as critical. So I went with a simple solution as a starting point. Definitely worth instrumenting the wait times though to validate that assumption.

src/main/java/com/salesforce/dynamodbv2/mt/util/RecordsCache.java
```diff
@@ -0,0 +1,276 @@
...
+            }
+
+            if (start.equals(from)) {
+                return records;
+            }
+
+            return records.subList(getIndex(from), records.size());
+        }
+
+        private int getIndex(SequenceNumber sequenceNumber) {
```
busjaeger: This is another behavior change (though non-functional) compared to master: instead of iterating over all records to find the index of a sequence number in a list of records, this impl uses binary search + lazy parsing of string sequence numbers into BigIntegers to allow comparison. I am hoping that will reduce the CPU load on our streams dynos.

src/main/java/com/salesforce/dynamodbv2/mt/util/RecordsCache.java
```diff
@@ -0,0 +1,276 @@
...
+        boolean isEmpty() {
+            return start.equals(end);
+        }
+    }
+
+    // config parameter
+    private final long maxRecordsByteSize;
+
+    // cached record segments sorted by sequence number within each shard
+    private final Map<ShardId, NavigableMap<SequenceNumber, Segment>> segments;
```
busjaeger: This is another behavior change (non-functional) compared to master: instead of using a single sorted map where the key class includes streamArn, shardId, and SequenceNumber, this impl uses a hashmap on streamArn+shardId and a nested sorted map on sequenceNumber. That should reduce the amount of computation needed to compare and lookup segments.

src/main/java/com/salesforce/dynamodbv2/mt/util/RecordsCache.java
```diff
@@ -0,0 +1,276 @@
...
+    // Insertion order of cache segments for eviction purposes
+    private final LinkedList<ShardLocation> insertionOrder;
```
busjaeger: In main this is a `LinkedHashMap`, since we are removing specific entries when merging segments. That's not needed anymore, since we only remove entries during eviction. If we bring back merging, we'll need to bring back that map as well.

src/main/java/com/salesforce/dynamodbv2/mt/util/RecordsCache.java
busjaeger: Great suggestion. The same holds for the shard-level lock (assuming we keep that).

src/main/java/com/salesforce/dynamodbv2/mt/util/SequenceNumber.java
```diff
@@ -0,0 +1,56 @@
+package com.salesforce.dynamodbv2.mt.util;
+
+import com.amazonaws.services.dynamodbv2.model.Record;
+import java.math.BigInteger;
+import java.util.Objects;
+
+final class SequenceNumber implements Comparable<SequenceNumber> {
```
busjaeger: Probably worth getting rid of this class and just using BigInteger directly instead. (Originally, this class was storing more stuff, but I got rid of that.)

src/main/java/com/salesforce/dynamodbv2/mt/util/ShardId.java
```diff
@@ -0,0 +1,57 @@
...
+final class ShardId {
+
+    @Nonnull
+    private final String streamArn;
+    @Nonnull
+    private final String shardId;
+
+    ShardId(String streamArn, String shardId) {
+        this.streamArn = checkNotNull(streamArn);
+        this.shardId = checkNotNull(shardId);
```
busjaeger: Could be worth pre-computing or caching the hashCode.

src/main/java/com/salesforce/dynamodbv2/mt/util/CachingAmazonDynamoDbStreams.java
msgroi: I have exposed the 2 caches so can be provided by the client that is created the MT AmazonDynamoDB via the SharedTableBuilder.  I haven't defined an interface that calls back with metrics.  But I have on my personal list a desire to record consumed capacities and that will necessitate such an interface so I will likely get to it soon.

src/main/java/com/salesforce/dynamodbv2/mt/util/CachingAmazonDynamoDbStreams.java
msgroi: 👍

src/main/java/com/salesforce/dynamodbv2/mt/util/StreamsCache.java
```diff
@@ -0,0 +1,398 @@
...
+    private final long maxRecordsByteSize;
+
+    // cached record segments sorted by sequence number within each shard
+    private final ConcurrentMap<ShardId, NavigableMap<BigInteger, Segment>> segments;
+    // Insertion order of cache segments for eviction purposes
+    private final Queue<ShardIteratorPosition> insertionOrder;
+    // locks for accessing shard caches
+    private final Striped<ReadWriteLock> shardLocks;
+    // size of cache >= 0
+    private final AtomicLong recordsByteSize;
```
busjaeger: @msgroi @shirdoo FYI, based on your feedback, I have made the cache lock more fine-grained (at the granularity of a shard). Would be good to have you sanity-check that the desired invariants still hold under concurrent access. The main difference compared to the previous version (which was using a single lock across all shards) is that eviction is done after the lock is released.

PR #438, opened 2019-07-02
Metrics

busjaeger:


busjaeger:
I am also working on adding metrics to the multitenant layer above the cache, but this is ready to be merged as-is.

PR #444, opened 2019-07-04
Update CHANGELOG.md

msgroi:


CHANGELOG.md
```diff
@@ -5,6 +5,10 @@ Multitenant AWS Dynamo supports the [AWS Dynamo Java API](http://docs.aws.amazon
   
 You can write your application code against the Amazon DynamoDB interface as you would for any other application.  The implementation will manage storage of data by tenant.
 
+## 0.9.42 (July 3, 2019)
+
+* Stream metrics refinements.
```
busjaeger: Agree with your suggestion: we should say something about adding micrometer and stream metrics. For some reason I thought we had already released an mt-dynamo version with that in it, because I have been working with the snapshot for a while locally.

PR #445, opened 2019-07-04
Add caching for describe stream

degaydorus:
Caches DescribeStreamResult (all shards) for given key (stream arn - physical arn).

Notes:
- The client should catch the AmazonDynamoDBException on describeStream (which our client is currently doing).
- Eviction happens on time as well as weight (large number of shards per cache entry)
- Metrics will be added in a future PR


src/test/java/com/salesforce/dynamodbv2/mt/mappers/MtAmazonDynamoDbStreamsBaseTest.java
```diff
@@ -422,4 +464,34 @@ void testGap(MtAmazonDynamoDbBase mtDynamoDb, MtAmazonDynamoDbStreams mtDynamoDb
...
+        try {
+            createTenantTables(mtDynamoDb);
+            List<Stream> streams = mtDynamoDbStreams.listStreams(new ListStreamsRequest()).getStreams();
+
+            for (Stream stream : streams) {
+                String key = stream.getStreamArn();
+                DescribeStreamResult expectedResult =
+                    new DescribeStreamResult().withStreamDescription(new StreamDescription().withStreamArn(key));
+
+                assertDescribeStreamCache(mtDynamoDbStreams, cachingStreams, stream, false, expectedResult);
```
degaydorus: This fails when run with all the tests, not when the test class is run.   I'm missing how the shardIterator might be empty in this case.  (looks like another test suite is interfering but I wasn't able to find which one). 
[ERROR] Failures:
[ERROR]   MtAmazonDynamoDbStreamsBaseTest.testDescribeStreamCache:490->assertDescribeStreamCache:250 expected: <true> but was: <false>
[ERROR]   MtAmazonDynamoDbStreamsBaseTest.testDescribeStreamCache:490->assertDescribeStreamCache:250 expected: <true> but was: <false>
[ERROR]   MtAmazonDynamoDbStreamsByTableTest>MtAmazonDynamoDbStreamsBaseTest.testDescribeStreamCache:490->MtAmazonDynamoDbStreamsBaseTest.assertDescribeStreamCache:250 expected: <true> but was: <false>
[ERROR]   MtAmazonDynamoDbStreamsByTableTest>MtAmazonDynamoDbStreamsBaseTest.testDescribeStreamCache:490->MtAmazonDynamoDbStreamsBaseTest.assertDescribeStreamCache:250 expected: <true> but was: <false>

src/main/java/com/salesforce/dynamodbv2/mt/util/CachingAmazonDynamoDbStreams.java
```diff
@@ -529,6 +568,96 @@ private CachingAmazonDynamoDbStreams(AmazonDynamoDBStreams amazonDynamoDbStreams
...
+            StreamDescription streamDescription = result.getStreamDescription();
+            streamDescription.setShards(allShards);
+            result.setStreamDescription(streamDescription);
+        } else if (previousResult != null) {
+            StreamDescription streamDescription = previousResult.getStreamDescription();
+            streamDescription.setShards(allShards);
+            result = previousResult;
+            result.setStreamDescription(streamDescription);
+        } else {
+            result = new DescribeStreamResult().withStreamDescription(new StreamDescription().withStreamArn(streamArn)
```
degaydorus: Wasn't sure if we want to return null on null `DescribeStreamResult` responses.  Right now we don't handle null case responses, probably because we don't expect it from the aws DescribeStream call.  Here I just return a basic stream description result.

src/main/java/com/salesforce/dynamodbv2/mt/util/CachingAmazonDynamoDbStreams.java
```diff
@@ -529,6 +568,96 @@ private CachingAmazonDynamoDbStreams(AmazonDynamoDBStreams amazonDynamoDbStreams
...
+        // TODO metrics: count cache miss
+
+        DescribeStreamResult result = null;
+        DescribeStreamResult previousResult;
+        do {
+            request = new DescribeStreamRequest().withStreamArn(streamArn).withExclusiveStartShardId(lastShardId);
+            previousResult = result;
+            result = super.describeStream(request);
+
+            if (result != null && result.getStreamDescription().getShards() != null) {
```
busjaeger: Is result ever null?

src/test/java/com/salesforce/dynamodbv2/testsupport/StreamsTestUtil.java
```diff
@@ -36,4 +42,30 @@ public static String mockSequenceNumber(int sequenceNumber) {
...
+        if (!expectedCacheHit) {
+            assertNull(cacheLookupResult);
+        } else {
+            System.out.println("cache lookup result: " + cacheLookupResult);
+            assertNotNull(cacheLookupResult);
+
+            assertTrue(cacheLookupResult.getStreamDescription().getStreamArn()
+                .equals(expectedResult.getStreamDescription().getStreamArn()));
+
+            if (expectedResult.getStreamDescription().getShards() != null) {
```
msgroi: Do you ever test the case where `expectedResult.getStreamDescription().getShards()` is null?

src/main/java/com/salesforce/dynamodbv2/mt/util/CachingAmazonDynamoDbStreams.java
```diff
@@ -487,6 +512,13 @@ private static String toShortString(GetRecordsResult result) {
     // iterator cache
     private final LoadingCache<CachingShardIterator, String> iteratorCache;
 
+    private Cache<String, DescribeStreamResult> describeStreamCache;
+
+    @VisibleForTesting
+    public Cache<String, DescribeStreamResult> getDescribeStreamCache() {
+        return describeStreamCache;
+    }
```
busjaeger: As discussed, I'd prefer to not expose the internal cache. Some possible strategies for testing could be:
- introduce builder/constructor variant that allows passing in cache reference
- assert cache hits/misses by observing calls to the passed in AmazonDynamoDBStreams instance

We can address this in a separate PR if we think it's worth doing.

src/main/java/com/salesforce/dynamodbv2/mt/util/CachingAmazonDynamoDbStreams.java
```diff
@@ -529,6 +568,97 @@ private CachingAmazonDynamoDbStreams(AmazonDynamoDBStreams amazonDynamoDbStreams
...
+     *
+     * @param describeStreamRequest Describe stream request.
+     * @return {@code DescribeStreamResult}.
+     */
+    @Override
+    public DescribeStreamResult describeStream(DescribeStreamRequest describeStreamRequest) {
+        if (LOG.isDebugEnabled()) {
+            LOG.debug("describeStream request={}", describeStreamRequest);
+        }
+        DescribeStreamResult result = lookupDescribeStreamResult(describeStreamRequest);
```
busjaeger: What is the value in having separate `describeStream` and `lookupDescribeStreamResult` methods?

src/main/java/com/salesforce/dynamodbv2/mt/util/CachingAmazonDynamoDbStreams.java
```diff
@@ -529,6 +568,97 @@ private CachingAmazonDynamoDbStreams(AmazonDynamoDBStreams amazonDynamoDbStreams
...
+     * @param describeStreamRequest Describe stream request.
+     * @return Stream details for the given request (streamArn).
+     */
+    private DescribeStreamResult lookupDescribeStreamResult(DescribeStreamRequest describeStreamRequest)
+        throws AmazonDynamoDBException {
+
+        String key = describeStreamRequest.getStreamArn();
+        try {
+            DescribeStreamResult result = describeStreamCache.get(key,
+                () -> this.loadStreamDescriptionForAllShards(describeStreamRequest));
```
busjaeger: Would it make sense to configure the loading method in when creating the cache, similar to what we do for the iterator cache?

src/main/java/com/salesforce/dynamodbv2/mt/util/CachingAmazonDynamoDbStreams.java
```diff
@@ -529,6 +568,97 @@ private CachingAmazonDynamoDbStreams(AmazonDynamoDBStreams amazonDynamoDbStreams
...
+            return result;
+        } catch (UncheckedExecutionException | ExecutionException | InvalidCacheLoadException e) {
+            // Catch exceptions thrown on cache lookup or cache loader and try load method once more.
+            // (get call will throw UncheckedExecutionException before aws exception (i.e. AmazonDynamoDBException)).
+            LOG.warn("Failed getting DescribeStreamResult for all shards from cache lookup or load method.  "
+                + "Retrying describeStream call.");
+            e.printStackTrace();
```
busjaeger: include log message in log message rather than printing stack trace?

src/main/java/com/salesforce/dynamodbv2/mt/util/CachingAmazonDynamoDbStreams.java
```diff
@@ -529,6 +568,97 @@ private CachingAmazonDynamoDbStreams(AmazonDynamoDBStreams amazonDynamoDbStreams
...
+                allShards.addAll(shards);
+                lastShardId = result.getStreamDescription().getLastEvaluatedShardId();
+            } else {
+                // since the request was null, update lastShardId to null since we can't get the next request
+                // from a null response
+                lastShardId = null;
+            }
+        } while (lastShardId != null);
+
+        if (result != null) {
```
busjaeger: Why not just create a new `DescribeStreamResult` at the end from whatever was loaded?

src/test/java/com/salesforce/dynamodbv2/mt/mappers/MtAmazonDynamoDbStreamsBaseTest.java
```diff
@@ -266,13 +86,50 @@ private static void assertGetRecords(MtAmazonDynamoDbStreams streams, Collection
...
+                                           CachingAmazonDynamoDbStreams cachingStreams,
+                                           Stream stream,
+                                           boolean expectedCacheHit,
+                                           DescribeStreamResult expectedResult) {
+        // Setup
+        String key = stream.getStreamArn();
+        Cache<String, DescribeStreamResult> describeStreamCache = cachingStreams.getDescribeStreamCache();
+
+        // Verify cache hit/miss based on expected result
+        StreamsTestUtil.verifyDescribeStreamCacheResult(describeStreamCache, key, expectedCacheHit, expectedResult);
```
busjaeger: Could we replace this method with a simple `assertEquals` on what is expected to be in the cache?

src/test/java/com/salesforce/dynamodbv2/mt/mappers/MtAmazonDynamoDbStreamsBaseTest.java
```diff
@@ -266,13 +86,50 @@ private static void assertGetRecords(MtAmazonDynamoDbStreams streams, Collection
...
+
+        // Trigger flow that interacts with the describe stream cache
+        Optional<String> shardIterator = BASE_TEST_UTILS.getShardIterator(mtDynamoDbStreams, stream);
+        assertTrue(shardIterator.isPresent());
```
busjaeger: Why assert that we can get a shard iterator? That seems beyond the scope of this test?

src/test/java/com/salesforce/dynamodbv2/mt/util/CachingAmazonDynamoDbStreamsTest.java
```diff
@@ -1033,4 +1110,265 @@ void testNonexistentShard() {
...
+        // The RateLimitingException exception is thrown twice since the cache loader method call (part of cache get)
+        // throws UncheckedExecutionException on AWS exceptions.  The retry will propagate the AmazonDynamoDBException.
+        when(mockStreams.describeStream(any())).thenThrow(expectedException, expectedException);
+
+        try {
+            cachingStreams.describeStream(new DescribeStreamRequest().withStreamArn(streamArn));
+            Assertions.fail();
+        } catch (AmazonDynamoDBException e) {
+            assertNotNull(e);
+        }
```
busjaeger: Nice coverage!

src/main/java/com/salesforce/dynamodbv2/mt/util/CachingAmazonDynamoDbStreams.java
degaydorus: In Kinesis Proxy when calling `describeStream`, it throws an exception if it's null after retrying and getting a null response: https://github.com/awslabs/amazon-kinesis-client/blob/09ec52bd16e5293cb06e53c05089b76adb30a76f/src/main/java/com/amazonaws/services/kinesis/clientlibrary/proxies/KinesisProxy.java#L276

I don't think the result will be null though by looking through the aws sdk.  If there's an error it should throw the known exceptions (which mt-dynamo clients should expect) - http://docs.aws.amazon.com/goto/WebAPI/kinesis-2013-12-02/DescribeStream.  Looking through the aws sdk, it looks like it returns a non null response since it calls response.getAwsResponse(); and adds an errorResponseHandler to make the response non-null if there is an error.

src/test/java/com/salesforce/dynamodbv2/mt/mappers/MtAmazonDynamoDbStreamsBaseTest.java
degaydorus: fixed in 8af378f4530cba778d8315f80d5b910073904e7a

src/test/java/com/salesforce/dynamodbv2/testsupport/StreamsTestUtil.java
degaydorus: added tests in f3d1d9ef8aec9f343ef093ee040b2fe75c8f169c

src/main/java/com/salesforce/dynamodbv2/mt/util/CachingAmazonDynamoDbStreams.java
degaydorus: updated in f3d1d9ef8aec9f343ef093ee040b2fe75c8f169c

src/main/java/com/salesforce/dynamodbv2/mt/util/CachingAmazonDynamoDbStreams.java
degaydorus: simplified in beed68d0c791549292ac4b01bf75b5b860af4ad6

src/main/java/com/salesforce/dynamodbv2/mt/util/CachingAmazonDynamoDbStreams.java
degaydorus: For consistency?

src/main/java/com/salesforce/dynamodbv2/mt/util/CachingAmazonDynamoDbStreams.java
degaydorus: updated in beed68d0c791549292ac4b01bf75b5b860af4ad6

src/main/java/com/salesforce/dynamodbv2/mt/util/CachingAmazonDynamoDbStreams.java
degaydorus: After removing the null checks, I first changed it to this:
```
result.getStreamDescription().setShards(allShards);
result.setStreamDescription(result.getStreamDescription());
```

then to this
```
new DescribeStreamResult().withStreamDescription(result.getStreamDescription().withShards(allShards));
```

I think either way is fine.

src/test/java/com/salesforce/dynamodbv2/mt/mappers/MtAmazonDynamoDbStreamsBaseTest.java
degaydorus: Not in this case because there are only parts of the StreamDescription we care to verify so this leads to multiple assertions (streamArn, shards, streamStatus)

For example these `Stream Description` fields are not affected by the describeStream caching code:
StreamViewType: NEW_AND_OLD_IMAGES,
CreationRequestDateTime: Tue Jul 09 18:21:14 PDT 2019,
TableName: MtAmazonDynamoDbStreamsBaseTest.testDescribeStreamCache.SharedTable_b,
KeySchema: [{AttributeName: id,KeyType: HASH}],

The `MtAmazonDynamoDbStreamsBaseTest` test will have a stream description that has all of these fields set.  The `verifyDescribeStreamCacheResult` method will assert expected values match for these fields: streamArn, shards, streamStatus.

I also think adding the checks if the result is null (cache hit/miss) adds necessary coverage.

src/test/java/com/salesforce/dynamodbv2/mt/mappers/MtAmazonDynamoDbStreamsBaseTest.java
degaydorus: Because the `shardIterator` gets used to verify the shard id's match a couple lines below:
https://github.com/salesforce/mt-dynamo/blob/cb16e79927723ba9641d53ff1ea6ca27bfcef740/src/test/java/com/salesforce/dynamodbv2/mt/mappers/MtAmazonDynamoDbStreamsBaseTest.java#L120

src/test/java/com/salesforce/dynamodbv2/mt/util/CachingAmazonDynamoDbStreamsTest.java
degaydorus: Thanks!

src/main/java/com/salesforce/dynamodbv2/mt/util/CachingAmazonDynamoDbStreams.java
```diff
@@ -529,6 +568,77 @@ private CachingAmazonDynamoDbStreams(AmazonDynamoDBStreams amazonDynamoDbStreams
...
+
+            if (result.getStreamDescription().getShards() != null
+                && !result.getStreamDescription().getShards().isEmpty()) {
+                List<Shard> shards = result.getStreamDescription().getShards();
+                allShards.addAll(shards);
+                lastShardId = result.getStreamDescription().getLastEvaluatedShardId();
+            } else { // If the shards list is null/empty exit since there shouldn't be more shards.
+                lastShardId = null;
+                result.getStreamDescription().setLastEvaluatedShardId(null);
+            }
```
busjaeger: I am not sure this assumption is correct. The docs say: “If LastEvaluatedShardId is not empty, it does not necessarily mean that there is more data in the result set. The only way to know when you have reached the end of the result set is when LastEvaluatedShardId is empty.“

So I think you need to keep going until you get a page that has an empty LastEvaluatedShardId regardless of whether there are shards in the list.

src/main/java/com/salesforce/dynamodbv2/mt/util/CachingAmazonDynamoDbStreams.java
```diff
@@ -81,13 +91,16 @@
         private static final int DEFAULT_MAX_GET_RECORDS_RETRIES = 10;
         private static final long DEFAULT_GET_RECORDS_LIMIT_EXCEEDED_BACKOFF_IN_MILLIS = 1000L;
         private static final int DEFAULT_MAX_ITERATOR_CACHE_SIZE = 100;
+        private static final long DEFAULT_DESCRIBE_STREAM_CACHE_TTL = 1;
```
busjaeger: Should we default to a slightly higher value (like 5s)?

degaydorus: updated in a032f5ee47ba57b46c150adadbf3552a1458a2af
Added a metric TODO for describeStream totalTime as this will help us adjust the default.

src/main/java/com/salesforce/dynamodbv2/mt/util/CachingAmazonDynamoDbStreams.java
degaydorus: Good catch.   Updated it to stop processing if LastEvaluatedShardId is null or empty.
impl and tests updated a032f5ee47ba57b46c150adadbf3552a1458a2af and 4114efc6a9911363f653a86e60abe9d0175f22c4

src/main/java/com/salesforce/dynamodbv2/mt/util/CachingAmazonDynamoDbStreams.java
```diff
@@ -175,12 +175,17 @@ public Builder withMaxIteratorCacheSize(int maxIteratorCacheSize) {
...
          * @return this Builder.
          */
-        public Builder withDescribeStreamCacheDisabled() {
-            LOG.info("describeStream cache is disabled");
-            this.describeStreamCacheEnabled = false;
+        public Builder withDescribeStreamCacheEnabled(boolean describeStreamCacheEnabled) {
+            this.describeStreamCacheEnabled = describeStreamCacheEnabled;
+
+            if (!this.describeStreamCacheEnabled) {
+                LOG.info("describeStream cache is disabled");
```
busjaeger: IMHO, this doesn't belong in the builder, but rather in code calling it (since this class is more of a generic utility and some callers may not want to log this or log it differently).

degaydorus: okay I'll take it out

PR #447, opened 2019-07-04
Misc cleanup 190704

horwitzj:


src/main/java/com/salesforce/dynamodbv2/mt/util/StreamsRecordCache.java
```diff
@@ -190,7 +190,9 @@ Segment subSegment(BigInteger from, BigInteger to) {
          * @return Index in list for given sequence number.
          */
         private int getIndex(BigInteger sequenceNumber) {
-            final List<BigInteger> sequenceNumbers = Lists.transform(records, StreamShardPosition::at);
+            final List<BigInteger> sequenceNumbers = records.stream()
```
horwitzj: Guava encourages migrating off of `Lists.transform`

busjaeger: I intentionally used Lists.transform here, because it’s lazy, so only sequence numbers that binary search looks at are parsed into BigIntegers. With this change we iterate and parse all entries.

.idea/checkstyle-idea.xml
```diff
@@ -3,12 +3,12 @@
   <component name="CheckStyle-IDEA">
     <option name="configuration">
       <map>
-        <entry key="active-configuration" value="LOCAL_FILE:$PRJ_DIR$/src/test/resources/checkstyle.xml:mt-dynamo-checks" />
-        <entry key="checkstyle-version" value="8.18" />
+        <entry key="active-configuration" value="LOCAL_FILE:$PROJECT_DIR$/src/test/resources/checkstyle.xml:mt-dynamo-checks" />
+        <entry key="checkstyle-version" value="8.22" />
```
msgroi: I have no idea whether `$PRJ_DIR` is better or worse than `$PROJECT_DIR`.

PR #449, opened 2019-07-08
initial metadata dump for backup-restore pipeline

sbabu-salesforce:


src/main/java/com/salesforce/dynamodbv2/mt/repo/MtDynamoDbTableDescriptionRepo.java
```diff
@@ -145,6 +155,33 @@ public TableDescription deleteTable(String tableName) {
...
+            .collect(Collectors.toMap(
+                rowItem ->
+                    getTenantTableFromHashKey(rowItem.get(tableDescriptionTableHashKeyField).getS())
+                ,
+                rowItem -> {
+                    String tableDataJson = rowItem.get(tableDescriptionTableDataField).getS();
+                    return getCreateTableRequest(jsonToTableData(tableDataJson));
+                })));
+
+        if (scanResult.getCount() == scanReq.getLimit()) {
```
sbabu-salesforce: will need to get to this obviously, but this will work until I try backing up a table with more tables than are returned in a single scan request

src/main/java/com/salesforce/dynamodbv2/mt/repo/MtDynamoDbTableDescriptionRepo.java
```diff
@@ -175,28 +212,67 @@ private void createTableDescriptionTableIfNotExists(int pollIntervalSeconds) {
...
-            long readCapacityUnits = capacityUtil.getCapacity(
-                    createTableRequest.getProvisionedThroughput(), DynamoDbCapacity.CapacityType.READ);
-            long writeCapacityUnits = capacityUtil.getCapacity(
-                    createTableRequest.getProvisionedThroughput(), DynamoDbCapacity.CapacityType.WRITE);
-
-            tableDescription.withProvisionedThroughput(new ProvisionedThroughputDescription()
-                    .withReadCapacityUnits(readCapacityUnits)
-                    .withWriteCapacityUnits(writeCapacityUnits));
-        }
+        // // Only set provisioned throughput if the BillingMode is not PAY_PER_REQUEST
```
sbabu-salesforce: I commented this provisioned throughput business out to get feedback from y'all, but I fundamentally disagree with where these defaults are being applied. 

IMO, we should take the exact CreateTableRequest the customer is giving us, and storing that in our metadata table. If we're operating against a shared table mt-dynamo, or we're operating against a pay per request table, we should ignore any provisioned throughput the customer specifies, but if we set defaults here, and migrate a tenant from a pay-per-request instance over to a table-per-tenant instance without pay-per-request for example, we lose any configuration the customer configured, same story for vice versa.

src/main/java/com/salesforce/dynamodbv2/mt/mappers/MtAmazonDynamoDbByTable.java
```diff
@@ -77,8 +77,6 @@ private MtAmazonDynamoDbByTable(MtAmazonDynamoDbContextProvider mtContext, Amazo
     }
 
     /**
-     * Determines if the table for the given name is a multitenant table associated with this instance.
```
horwitzj: any reason to delete this doc (or the yet-to-be-fleshed-out docs below)?

src/main/java/com/salesforce/dynamodbv2/mt/repo/MtTableDescriptionRepo.java
```diff
@@ -9,18 +9,41 @@
...
- * TODO: write Javadoc.
+ * Interface for managing table metadata of multi-tenant virtual tables within mt-dynamo.
  *
  * @author msgroi
  */
 public interface MtTableDescriptionRepo {
 
+    /**
+     * Create a multi-tenant (virtual) table with specs defined in @param createTableRequest under the given
+     * multi-tenant contexts' namespace.
```
horwitzj: assuming this refers to just one context, replace "contexts'" with "context's"
also, does "@param" just show up as "@param" in the Javadoc?

src/test/resources/checkstyle.xml
```diff
@@ -235,10 +235,6 @@ Change log:
         </module>
         <module name="NonEmptyAtclauseDescription"/>
         <module name="JavadocTagContinuationIndentation"/>
-        <module name="SummaryJavadoc">
```
horwitzj: why remove this?

src/main/java/com/salesforce/dynamodbv2/mt/mappers/MtAmazonDynamoDb.java
```diff
@@ -95,4 +96,40 @@ public String toString() {
...
+        @Override
+        public boolean equals(Object o) {
+            if (this == o) {
+                return true;
+            }
+            if (o == null || getClass() != o.getClass()) {
+                return false;
+            }
+            TenantTable that = (TenantTable) o;
+            return virtualTable.equals(that.virtualTable) &&
```
horwitzj: move "&&" to next line

src/main/java/com/salesforce/dynamodbv2/mt/mappers/MtAmazonDynamoDbByTable.java
sbabu-salesforce: this doc is [almost] literally identical to the snippet after @return, useless IMO. 

Also, the javadocs `TODO: write Javadoc.` bothers/annoys me almost physically, maybe that's my OCD.

src/test/resources/checkstyle.xml
sbabu-salesforce: I find for many methods, this leads to redundant code. For basic functions, a javadoc 
```
@returns an object transformed from input with blah blah behavior
```

seems totally sufficient. By requiring both a description and a returns flag, there is often comment duplication. For functions doing interesting things, sure add both a description, and param/return snippets, but don't enforce it an all methods is my opinion.

src/test/resources/checkstyle.xml
```diff
@@ -235,11 +235,6 @@ Change log:
...
-            <property name="forbiddenSummaryFragments"
-              value="^@return the *|^This method returns |^A [{]@code [a-zA-Z0-9]+[}]( is a )"/>
-        </module>
-        <module name="JavadocParagraph"/>
```
sbabu-salesforce: Also intentional. I don't agree with requiring prefixing paragraphs in javadocs with `<p>` as adding value, instead, I think tedious requirements like that actually dissuade terse javadoc additions, and instead, leads to writing functions with empty javadocs, to "document in a later PR", which also means "never"

src/main/kotlin/com/salesforce/dynamodbv2/mt/repo/MtTableDescriptionRepo.kt
```diff
@@ -0,0 +1,67 @@
...
+     * TODO: these overloaded listVirtualTableMetadatas are placeholders from implementing this interface in Java.
+     * (see https://stackoverflow.com/questions/47070968/kotlin-default-arguments-in-interface-bug)
+     * Once the impl is converted to Kotlin, all the below signatures can be removed with defaulted params.
+     */
+    fun listVirtualTableMetadatas(exclusiveStartTableMetadata: TenantTableMetadata?): ListMetadataResult
+    fun listVirtualTableMetadatas(limit: Int): ListMetadataResult
+    fun listVirtualTableMetadatas(): ListMetadataResult
+    data class ListMetadataResult(val metadataList: List<TenantTableMetadata>, val lastEvaluatedTable: TenantTableMetadata?)
+
+    data class TenantTableMetadata(val tenantTable: TenantTable, val createTableRequest: CreateTableRequest)
```
busjaeger: I wonder if it would be more consistent with other interfaces (like `scan` and `getRecords`) to return a list of `CreateTableRequests` that contain the context (maybe an `MtCreateTableRequest`). This class threw me off at first, because the `tenantTable` field contains the `virtualTableName` which the same as the `tableName` in the `CreateTableRequest`, correct? IIUC, this is an internal interface atm, so we could revisit in the future?

src/main/kotlin/com/salesforce/dynamodbv2/mt/sample/SampleDataClass.kt
```diff
@@ -0,0 +1,11 @@
...
+ * All rights reserved.
+ * SPDX-License-Identifier: BSD-3-Clause.
+ * For full license text, see LICENSE file in the repo root or https://opensource.org/licenses/BSD-3-Clause.
+ */
+package com.salesforce.dynamodbv2.mt.sample
+
+/**
+ * just a demo that kotlin has been assimilated into the build just fine
+ */
+data class SampleDataClass(val foo: String = "bar")
```
busjaeger: Do we still need this, given `MtTableDescriptionRepo.kt` above?

src/main/java/com/salesforce/dynamodbv2/mt/mappers/sharedtable/impl/RecordMapper.java
```diff
@@ -40,14 +40,31 @@
...
-        return new MtRecord()
-            .withAwsRegion(record.getAwsRegion())
-            .withEventID(record.getEventID())
-            .withEventName(record.getEventName())
-            .withEventSource(record.getEventSource())
-            .withEventVersion(record.getEventVersion())
-            .withUserIdentity(record.getUserIdentity())
-            .withContext(mtContext.getContext())
+        MtRecord ret = new MtRecord();
+        if (record.getAwsRegion() != null) {
```
busjaeger: What's the reason for this change? I believe most of these fields usually are not null, so now we're calling each getter twice.

src/main/java/com/salesforce/dynamodbv2/mt/repo/MtDynamoDbTableDescriptionRepo.java
```diff
@@ -145,6 +156,11 @@ public TableDescription deleteTable(String tableName) {
         return tableDescription;
     }
 
+    private TenantTable getTenantTableFromHashKey(String hashKey) {
```
busjaeger: Can we just use `TenantTable` as the map key? Using a string can cause collisions, so it's been on our todo list to replace it.

src/main/java/com/salesforce/dynamodbv2/mt/repo/MtDynamoDbTableDescriptionRepo.java
```diff
@@ -175,28 +191,59 @@ private void createTableDescriptionTableIfNotExists(int pollIntervalSeconds) {
             pollIntervalSeconds);
     }
 
+    private CreateTableRequest getCreateTableRequest(TableDescription description) {
```
busjaeger: Could these methods be static?

src/main/kotlin/com/salesforce/dynamodbv2/mt/repo/MtTableDescriptionRepo.kt
```diff
@@ -0,0 +1,67 @@
...
+    fun deleteTable(tableName: String): TableDescription
+
+    /**
+     * Utility to enumerate all virtual table metadata managed by this instance. Return up to @param limit results,
+     * starting after @param exclusiveStartTableMetadata if specified.
+     *
+     * @return a list of TenantTableMetadata objects, with a lastEvaluatedTable populated to the last table name if the
+     * result set fully populated @param limit results, null otherwise.
+     */
+    fun listVirtualTableMetadatas(
```
busjaeger: I think `metadata` is already plural, so may want to drop the `s`.

src/main/kotlin/com/salesforce/dynamodbv2/mt/repo/MtTableDescriptionRepo.kt
```diff
@@ -0,0 +1,67 @@
...
+        limit: Int,
+        exclusiveStartTableMetadata: TenantTableMetadata?
+    ): ListMetadataResult
+
+    /**
+     * TODO: these overloaded listVirtualTableMetadatas are placeholders from implementing this interface in Java.
+     * (see https://stackoverflow.com/questions/47070968/kotlin-default-arguments-in-interface-bug)
+     * Once the impl is converted to Kotlin, all the below signatures can be removed with defaulted params.
+     */
```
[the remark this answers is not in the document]
busjaeger: Reading the test code, I think using a request object may actually make it easier to see which variant is being called. For example, `listVirtualTableMetadatas(1, null)` would be `listVirtualTableMetadatas(new ListMetadataRequest().withLimit(1))`. More verbose, but more symmetric and aligned with DynamoDB API.

src/test/java/com/salesforce/dynamodbv2/mt/repo/MtDynamoDbTableDescriptionRepoTest.java
```diff
@@ -112,4 +118,49 @@ void testMtDynamoDbTableDescriptionPayPerRequestIsSet() throws InterruptedExcept
...
+        TenantTableMetadata tenantTable1 = new TenantTableMetadata(new TenantTable(tableName1, tenant1), createReq1);
+        TenantTableMetadata tenantTable2 = new TenantTableMetadata(new TenantTable(tableName2, tenant2), createReq2);
+        ListMetadataResult expected = new ListMetadataResult(
+            ImmutableList.of(tenantTable1, tenantTable2), null);
+
+        assertEquals(expected, returnedMetadatas);
+
+        returnedMetadatas = repo.listVirtualTableMetadatas(1, null);
+        expected = new ListMetadataResult(ImmutableList.of(tenantTable1), tenantTable1);
+        assertEquals(expected, returnedMetadatas);
```
busjaeger: In the interest of making test code easier to understand and having having one assert per test method, I'd pull these into separate test methods.

You may also want to add some test cases for boundary conditions, e.g., empty repo, specifying the max value for limit, etc.

sbabu-salesforce: makes sense, will submit a followup PR

src/main/kotlin/com/salesforce/dynamodbv2/mt/repo/MtTableDescriptionRepo.kt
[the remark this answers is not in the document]
sbabu-salesforce: ahh, ListMetadataRequest object is definitly preferable, I was stuck trying to get defaulted parameters with kotlin working, which would avoid all the below signatures, basically:
```
fun listVirtualTableMetadatas(
        limit: Int = 10,
        exclusiveStartTableMetadata: TenantTableMetadata? = null
    ): ListMetadataResult
```

I think given there's only two params, a request object is a little overkill, but that's much preferable over what we have here, with 4 extra signatures for every default combination, and one more parameter would blow this up to 8 (https://stackoverflow.com/questions/47070968/kotlin-default-arguments-in-interface-bug)

src/main/java/com/salesforce/dynamodbv2/mt/mappers/MtAmazonDynamoDbByTable.java
sbabu-salesforce: pulled the whole snippet out, already javadoc-ed in super class

src/main/java/com/salesforce/dynamodbv2/mt/mappers/sharedtable/impl/RecordMapper.java
sbabu-salesforce: so the reason for this is, we have tests creating mocked Record instances, so all these are nulled in said tests. 

Java has no problem letting these values be set to null, but when I converted to Kotlin, it set all the types to be Non-null, enforced at runtime. So we have three options: 
1) when translating to Kotlin, keep the lax nullable values
2) Update the test to go further with the mock and set default values (or use real Record instances)
3) enforce the non-null fields in MtRecord constructors and factories

This is option 3), but thinking about it, option 2) might be safer. Kind of kills the actual proper null handling of Kotlin though, if we let everything be nullable, sort of introducing tech debt to a solvable problem (that's much uglier/harder/prohibitively tedious to solve in Java)

sbabu-salesforce: advice on which of the above 3 approaches are preffered would be appreciated. Again, option 3) has been merged in, but I can see objections to muddying prod code to deal with test reqs (converting mock `Record` objects to real `MtRecord` instances.

src/main/java/com/salesforce/dynamodbv2/mt/repo/MtDynamoDbTableDescriptionRepo.java
sbabu-salesforce: not sure I follow... the hash key within the metadata table at the moment is a string (tenantName + delimeter + virtualTableName). The map key therefore is to take a row from Dynamo, and serialize the row with its `TenantTable` and `CreateTableRequest` onto S3 per tenant-table backup. 
Did you mean use `TenantTable` serialized into a binary hash key onto our dynamo metadata table?

src/main/java/com/salesforce/dynamodbv2/mt/mappers/sharedtable/impl/RecordMapper.java
busjaeger: If Kotlin is happy with what you have, then I am fine keeping it this way. Was just trying to understand the reason. If this is necessary, it may also be useful to add a comment about it. Also makes me wonder if this difference in null support could come up in other places.

src/main/java/com/salesforce/dynamodbv2/mt/repo/MtDynamoDbTableDescriptionRepo.java
busjaeger: Sorry, I got mixed up. In memory MtCache adds the prefix and we already changed that from string concatenation to a data class. The hash key for the DynamoDB table is still computed with unsafe string concatenation; to fix that we’d have to use something similar to what we’re doing for data tables (String/BinaryPrefixFunction). That may require dropping the metadata table, so we should tackle that separately.

src/main/kotlin/com/salesforce/dynamodbv2/mt/repo/MtTableDescriptionRepo.kt
sbabu-salesforce: makes sense, included that in https://github.com/salesforce/mt-dynamo/pull/476

src/test/java/com/salesforce/dynamodbv2/mt/repo/MtDynamoDbTableDescriptionRepoTest.java
sbabu-salesforce: included in https://github.com/salesforce/mt-dynamo/pull/476

PR #450, opened 2019-07-09
Optimize stream record filtering

busjaeger:


src/main/java/com/salesforce/dynamodbv2/mt/mappers/MtAmazonDynamoDbStreamsBase.java
```diff
@@ -143,32 +136,18 @@ public GetRecordsResult getRecords(GetRecordsRequest getRecordsRequest) {
             LOG.debug("getRecords request={}", getRecordsRequest);
```
busjaeger: `getRecords` base class behavior change: instead of assuming we always need to filter and map, just determine whether the method is called with tenant context or without. The actual retrieving of records is delegated to the sub-class (currently either byTable or byIndex).

src/main/java/com/salesforce/dynamodbv2/mt/mappers/MtAmazonDynamoDbStreamsBase.java
```diff
@@ -143,32 +136,18 @@ public GetRecordsResult getRecords(GetRecordsRequest getRecordsRequest) {
...
         }
 
-        ShardIterator iterator = ShardIterator.fromString(getRecordsRequest.getShardIterator());
-        String arn = iterator.getArn();
-        StreamArn streamArn = parse(arn);
+        final ShardIterator iterator = ShardIterator.fromString(getRecordsRequest.getShardIterator());
+        final String arn = iterator.getArn();
+        final StreamArn streamArn = parse(arn);
```
busjaeger: Parsing the ARN determines whether the method is called for a virtual table with tenant context or for a physical table without tenant context.

busjaeger: (`parse` validates that current thread context matches arn context.)

src/main/java/com/salesforce/dynamodbv2/mt/mappers/MtAmazonDynamoDbStreamsBase.java
```diff
@@ -143,32 +136,18 @@ public GetRecordsResult getRecords(GetRecordsRequest getRecordsRequest) {
...
-                    .map(StreamRecord::getSequenceNumber)
-                    .orElse(null));
-        }
+        final GetRecordsRequest request = new GetRecordsRequest()
+            .withShardIterator(iterator.withArn(streamArn.toDynamoDbArn()).toString())
+            .withLimit(getRecordsRequest.getLimit());
+
+        final MtGetRecordsResult result = streamArn instanceof MtStreamArn
+            ? getMtRecords(request, (MtStreamArn) streamArn)
+            : getMtRecords(request, streamArn);
```
busjaeger: Depending on parse result, dispatches to one of two methods implemented by subclasses.

src/main/java/com/salesforce/dynamodbv2/mt/mappers/MtAmazonDynamoDbStreamsByTable.java
```diff
@@ -17,9 +23,31 @@
     }
 
     @Override
-    protected Function<Record, MtRecord> getRecordMapper(StreamArn arn) {
-        String[] tenantAndTableName = mtDynamoDb.getTenantAndTableName(arn.getTableName());
-        return record -> mapRecord(tenantAndTableName[0], tenantAndTableName[1], record);
+    protected MtGetRecordsResult getMtRecords(GetRecordsRequest request, StreamArn streamArn) {
+        final String[] tenantAndTableName = mtDynamoDb.getTenantAndTableName(streamArn.getTableName());
+        return getMtRecords(request, tenantAndTableName[0], tenantAndTableName[1]);
+    }
```
busjaeger: If by table is called for physical table, it derives the tenant context and table name from the physical table.

src/main/java/com/salesforce/dynamodbv2/mt/mappers/MtAmazonDynamoDbStreamsByTable.java
```diff
@@ -17,9 +23,31 @@
...
+
+    @Override
+    protected MtGetRecordsResult getMtRecords(GetRecordsRequest request, MtStreamArn mtStreamArn) {
+        return getMtRecords(request, mtStreamArn.getContext(), mtStreamArn.getTenantTableName());
+    }
```
busjaeger: If it's called in tenant context, then we can use the parsed values from the arn directly.

Other than that, the methods do no differ for this strategy.

src/main/java/com/salesforce/dynamodbv2/mt/mappers/MtAmazonDynamoDbStreamsByTable.java
```diff
@@ -17,9 +23,31 @@
...
+        for (Record record : records) {
+            mtRecords.add(mapRecord(tenant, tenantTable, record));
+        }
+        return new MtGetRecordsResult()
+            .withRecords(mtRecords)
+            .withNextShardIterator(nextIterator)
+            .withLastSequenceNumber(getLast(mtRecords).getDynamodb().getSequenceNumber());
     }
 
     private MtRecord mapRecord(String tenant, String tableName, Record record) {
```
busjaeger: Mapping records for byTable is straightforward (and unchanged): only adds tenant id and table name to record (no field mapping needed).

src/main/java/com/salesforce/dynamodbv2/mt/mappers/sharedtable/impl/BinaryFieldMapper.java
```diff
@@ -38,6 +39,13 @@ public AttributeValue reverse(FieldMapping fieldMapping, AttributeValue qualifie
         return convertFromBinary(fieldMapping.getTarget().getType(), fieldValue.getValue());
     }
 
+    @Override
+    public Predicate<AttributeValue> createFilter() {
+        final Predicate<ByteBuffer> prefixFilter =
+            BinaryFieldPrefixFunction.INSTANCE.createFilter(mtContext.getContext(), virtualTableName);
+        return attributeValue -> prefixFilter.test(attributeValue.getB());
+    }
```
busjaeger: `FieldMapper`s and `PrefixFunction`s now support creating filters, so we can push the predicate as low as possible (this is the main perf optimization).

src/main/java/com/salesforce/dynamodbv2/mt/mappers/sharedtable/impl/BinaryFieldPrefixFunction.java
```diff
@@ -49,6 +44,13 @@ public ByteBuffer apply(FieldValue<ByteBuffer> fieldValue) {
         return new FieldValue<>(context, tableName, ByteBuffer.wrap(value));
     }
 
+    @Override
+    public Predicate<ByteBuffer> createFilter(String context, String tableName) {
+        final ByteBuffer prefix = newBuffer(context, tableName, 0).flip();
+        final int end = prefix.limit();
+        return b -> b.asReadOnlyBuffer().mismatch(prefix.asReadOnlyBuffer()) == end;
```
busjaeger: This is the equivalent of `String.startsWith` for binary hash keys.

src/main/java/com/salesforce/dynamodbv2/mt/mappers/sharedtable/impl/MtAmazonDynamoDbStreamsBySharedTable.java
```diff
@@ -55,6 +54,31 @@ public GetShardIteratorResult getShardIterator(GetShardIteratorRequest request)
         return super.getShardIterator(request);
     }
 
+    @Override
+    protected MtGetRecordsResult getMtRecords(GetRecordsRequest request, StreamArn streamArn) {
```
busjaeger: If we retrieve records without tenant context (KCL), then we don't need to filter anything, but we do need to use the prefix on each record to establish tenant context, so we can map each record using the virtual table definition.

src/main/java/com/salesforce/dynamodbv2/mt/mappers/sharedtable/impl/MtAmazonDynamoDbStreamsBySharedTable.java
```diff
@@ -66,8 +90,7 @@ public GetShardIteratorResult getShardIterator(GetShardIteratorRequest request)
      * stream shard, so that clients can request a new shard iterator where they left off.
      */
     @Override
-    protected MtGetRecordsResult getRecords(GetRecordsRequest request, Function<Record, MtRecord> recordMapper,
-                                            Predicate<MtRecord> recordFilter) {
+    protected MtGetRecordsResult getMtRecords(GetRecordsRequest request, MtStreamArn mtStreamArn) {
```
busjaeger: If we are called with tenant context, we need to filter records to those that have matching context, but then we can use a cached record mapper to convert them.

src/main/java/com/salesforce/dynamodbv2/mt/mappers/sharedtable/impl/MtAmazonDynamoDbStreamsBySharedTable.java
```diff
@@ -159,68 +187,20 @@ private int retryLoadRecords(MtGetRecordsResult mtResult,
...
         int consumedRecords = 0;
         for (Record record : records) {
-            MtRecord mtRecord = recordMapper.apply(record);
-            if (recordFilter.test(mtRecord)) {
+            if (recordFilter.test(record)) {
                 if (mtRecords.size() >= limit) {
                     break;
                 } else {
-                    mtRecords.add(mtRecord);
+                    mtRecords.add(recordMapper.apply(record));
```
busjaeger: This is the main optimization: instead of converting each record first and then filtering it, filter first and then convert (using a cached mapper rather than establishing tenant context for each record).

src/main/java/com/salesforce/dynamodbv2/mt/mappers/sharedtable/impl/RecordMapper.java
```diff
@@ -0,0 +1,57 @@
...
+
+import com.amazonaws.services.dynamodbv2.model.AttributeValue;
+import com.amazonaws.services.dynamodbv2.model.Record;
+import com.amazonaws.services.dynamodbv2.model.StreamRecord;
+import com.salesforce.dynamodbv2.mt.context.MtAmazonDynamoDbContextProvider;
+import com.salesforce.dynamodbv2.mt.mappers.MtAmazonDynamoDb.MtRecord;
+import java.util.function.Function;
+import java.util.function.Predicate;
+
+public class RecordMapper implements Function<Record, MtRecord> {
```
busjaeger: Moved record filtering and mapping logic here, so that this class can be cached as part of `TableMapping`.

src/main/java/com/salesforce/dynamodbv2/mt/mappers/sharedtable/impl/MtAmazonDynamoDbStreamsBySharedTable.java
```diff
@@ -159,68 +187,20 @@ private int retryLoadRecords(MtGetRecordsResult mtResult,
...
         }
         return consumedRecords;
     }
 
-    @Override
-    protected Predicate<MtRecord> getRecordFilter(StreamArn arn) {
-        Predicate<MtRecord> defaultPredicate = super.getRecordFilter(arn);
-        return mtRecord ->
-            mtDynamoDb.getMtContext().withContext(mtRecord.getContext(), this::isStreamEnabled, mtRecord.getTableName())
-                && defaultPredicate.test(mtRecord);
```
busjaeger: One intentional side-effect of this change is that `isStreamEnabled` is no longer called as part of the record filter.

For calling with context, we already check whether the stream is enabled when we return the virtual ARN (in `describeTable`). We could check again at the beginning of `getRecords`, but tables are currently immutable, so it seems unnecessary.

For calling without context (KCL), we have observed that clients actually need to see all records. For example, KCL needs to see them to set checkpoints to the latest record in the physical stream. @sbabu-salesforce had also brought up the need to see all records for backup/restore. It could be argued that this somewhat breaks the abstraction, since the client now has to check on its end whether the virtual table for a given record has streaming enabled. On the other hand, we are already breaking the abstraction by allowing calling without context, so hopefully that's fine.

src/main/java/com/salesforce/dynamodbv2/mt/mappers/MtAmazonDynamoDbStreamsBase.java
msgroi: Do we want to have a flag that gets passed into the `MtAmazonDynamoDbStreamsBase` that tells it that it should require context?  This would preclude accidentally falling back to reading all records from the stream accidently.

src/main/java/com/salesforce/dynamodbv2/mt/mappers/sharedtable/impl/MtAmazonDynamoDbStreamsBySharedTable.java
msgroi: Can we put some version of what you said above in the code comments?

src/main/java/com/salesforce/dynamodbv2/mt/cache/MtCache.java
```diff
@@ -25,19 +25,46 @@
  *
  * @author msgroi
  */
-public class MtCache<V> implements Cache<String, V> {
+public class MtCache<V> implements Cache<Object, V> {
+
+    private static class Key {
```
busjaeger: Note: this is needed both for performance and correctness (since current delimiter-based approach can result in collisions).

src/main/java/com/salesforce/dynamodbv2/mt/mappers/MtAmazonDynamoDbStreamsBase.java
```diff
@@ -36,6 +36,18 @@
 
     private static final Logger LOG = LoggerFactory.getLogger(MtAmazonDynamoDbStreamsBase.class);
 
+    /**
+     * Parses arn and validates that tenant context matches current thread tenant context.
+     *
+     * @param arn Arn to parse.
```
horwitzj: "Arn" -> "ARN"; also, this is not a sentence, so no terminal "." (i.e., "@param arn ARN to parse")

similarly elsewhere (e.g., "@return parsed ARN" on the next line)

busjaeger: I am putting periods at the end of everything, because checkstyle is making me paranoid and because we have Jeremy GC ;)

busjaeger: I'll fix the javadoc in a future PR (unless you beat me to it).

PR #453, opened 2019-07-10
Initial backup restore

sbabu-salesforce:


sbabu-salesforce:
The change looks large and scary, but I've been trying hard to keep most of the changes small and isolated. Most of the java changes that appear in this PR are captured in these 3 other PRs: 
1) Add kotlin to mt-dynamo: https://github.com/salesforce/mt-dynamo/pull/375
2) Add multi-tenant scan support to generate backup: https://github.com/salesforce/mt-dynamo/pull/411
3) Add virtual table metadata dump to attach to backup: https://github.com/salesforce/mt-dynamo/pull/449
 
I'm hoping to get those PR's merged first, at which point this should just be small changes with all the new Kotlin backup/restore code and test.

.idea/codeStyles/Project.xml
```diff
@@ -18,15 +18,18 @@
...
         <value>
           <package name="" withSubpackages="true" static="true" />
           <emptyLine />
           <package name="" withSubpackages="true" static="false" />
         </value>
       </option>
     </JavaCodeStyleSettings>
-  <codeStyleSettings language="JAVA">
+    <JetCodeStyleSettings>
+      <option name="NAME_COUNT_TO_USE_STAR_IMPORT" value="2147483647" />
```
sbabu-salesforce: intellij by default uses star imports, which ktlint does not like. This setting to intellij makes ktlint happy

Makefile
```diff
@@ -1,5 +1,5 @@
 build:
-	mvn test-compile -DskipTests=true -B -V
+	mvn spotless:check test-compile -DskipTests=true -B -V
```
sbabu-salesforce: got annoyed with catching style failures way late in the test process, so running this first to weed out silly stylechecks early in the build

Makefile
```diff
@@ -8,7 +8,7 @@ integration:
 	mvn -Dskip.surefire.tests -DskipITs=false failsafe:integration-test@failsafe-execution
 
 validate:
-	mvn javadoc:javadoc@javadoc-execution
+	#mvn javadoc:javadoc@javadoc-execution
```
sbabu-salesforce: this is something I should tack onto the kotlin PR, but unfortunately, the mvn javadoc plugin doesn't play well with kotlin. This article (https://kotlinlang.org/docs/reference/using-maven.html#generating-documentation) suggests using dokka? instead of the javadoc plugin with maven, haven't tried that out yet.

src/main/java/com/salesforce/dynamodbv2/mt/mappers/sharedtable/SharedTableBuilder.java
```diff
@@ -222,6 +225,11 @@ public SharedTableBuilder withGetRecordsTimeLimit(long getRecordsTimeLimit) {
         return this;
     }
 
+    public SharedTableBuilder withBackupSupport(String region, String backupS3BucketName) {
```
sbabu-salesforce: this is the hook for a client would configure backup support. Note, this implementation is very much tied to our shared table implementation, but I see no reason why we can't get this to work against the other two as well (byTable and byAccount), just more work to prioritize into the backlog.

src/main/java/com/salesforce/dynamodbv2/mt/mappers/sharedtable/impl/MtAmazonDynamoDbBySharedTable.java
```diff
@@ -497,6 +545,22 @@ public UpdateItemResult updateItem(UpdateItemRequest updateItemRequest) {
         return getAmazonDynamoDb().updateItem(updateItemRequest);
     }
 
+    @Override
+    public RestoreTableFromBackupResult restoreTableFromBackup(
+        RestoreTableFromBackupRequest restoreTableFromBackupRequest) {
+        throw new UnsupportedOperationException();
```
sbabu-salesforce: these are still throwing unsupported operation exceptions, so doing nothing at the moment, but these are the hooks I imagined clients of mt-dynamo will use to consume this backup/restore support. 

I figure, all we need is a translation of these dynamo restore/backup requests, to the simpler backup/restore request objects defined in `MtBackupManager.kt` and vice versa for the result objects. Something close to my next todos, but I figure, comes secondary to any review feedback of this and ancillary PRs

src/main/java/com/salesforce/dynamodbv2/mt/repo/MtTableDescriptionRepo.java
```diff
@@ -9,18 +9,41 @@
 
 import com.amazonaws.services.dynamodbv2.model.CreateTableRequest;
 import com.amazonaws.services.dynamodbv2.model.TableDescription;
+import com.salesforce.dynamodbv2.mt.sharedtable.impl.TenantTable;
+import java.util.Map;
 
 /**
- * TODO: write Javadoc.
+ * Interface for managing table metadata of multi-tenant virtual tables within mt-dynamo.
```
horwitzj: "multi-tenant" -> "multitenant"

src/main/kotlin/com/salesforce/dynamodbv2/mt/sharedtable/MtBackupManager.kt
```diff
@@ -0,0 +1,84 @@
...
+        mtContext: MtAmazonDynamoDbContextProvider
+    ): TenantRestoreMetadata
+
+    /**
+     * List all multi-tenant backups known to us on S3.
+     */
+    fun listMtBackups(): List<MtBackupMetadata>
+}
+
+data class MtBackupMetadata(val mtBackupId: String, val status: Status, val tenantTables: Set<TenantTableBackupMetadata>)
```
sbabu-salesforce: being able to define data classes like this without needing lombok is maybe one of my favorite parts of Kotlin. So much support for keeping code concise, with extensions for customizing equals or toString() or anything else you'd want to customize.

.idea/codeStyles/Project.xml
horwitzj: FWIW, wildcard imports can be disabled in Intellij: https://www.jetbrains.com/help/idea/creating-and-optimizing-imports.html#disable-wildcard-imports

Makefile
horwitzj: could the javadoc plugin run directory-based (or some other way to sort out .kt vs .java—it seems odd that the plug-in would even bother with .kt files)? (some option like `<dependencySourceExcludes>` (though maybe not that one specifically)?)

.idea/codeStyles/Project.xml
sbabu-salesforce: right, that's what I did which caused this diff on this intellij file. I wanna say this sets the default so anyone else building this repo will get the same setting, so everyone wont need to go through and disable wildcard imports manually to get ktlint to pass, at least that would be the ideal case

Makefile
sbabu-salesforce: unfortunately, its not failing when trying to process the Kotlin files, it fails when a Java class imports a Kotlin class, and fails with a package not found: 
```
[ERROR] Exit code: 1 - /Users/sbabu/Workspaces/mt-dynamo/src/main/java/com/salesforce/dynamodbv2/mt/mappers/MtAmazonDynamoDb.java:8: error: package com.salesforce.dynamodbv2.mt.sample does not exist
[ERROR] import com.salesforce.dynamodbv2.mt.sample.SampleDataClass;
[ERROR]                                           ^
[ERROR] /Users/sbabu/Workspaces/mt-dynamo/src/main/java/com/salesforce/dynamodbv2/mt/mappers/MtAmazonDynamoDb.java:16: error: cannot find symbol
[ERROR]     SampleDataClass getSample();
[ERROR]     ^
[ERROR]   symbol:   class SampleDataClass
[ERROR]   location: interface MtAmazonDynamoDb
```

Will try out dokka today

src/main/java/com/salesforce/dynamodbv2/mt/mappers/sharedtable/impl/MtAmazonDynamoDbBySharedTable.java
```diff
@@ -440,6 +463,31 @@ public ScanResult scan(ScanRequest scanRequest) {
         return scanResult;
     }
 
+    private ScanResult multiTenantScan(ScanRequest scanRequest) {
```
horwitzj: `multiTenant` -> `multitenant`

src/main/java/com/salesforce/dynamodbv2/mt/repo/MtDynamoDbTableDescriptionRepo.java
```diff
@@ -145,6 +155,33 @@ public TableDescription deleteTable(String tableName) {
         return tableDescription;
     }
 
+    @Override
+    public Map<TenantTable, CreateTableRequest> getAllMtTables() {
```
busjaeger: Should we considering returning a Stream or something similar from this method? I could imagine this collection getting too large to load into memory.

.idea/codeStyles/Project.xml
horwitzj: ooh that's great—thanks!

src/main/java/com/salesforce/dynamodbv2/mt/repo/MtDynamoDbTableDescriptionRepo.java
sbabu-salesforce: addressed in https://github.com/salesforce/mt-dynamo/pull/449

PR #454, opened 2019-07-10
Remove unnecessary ByteBuffer creation

busjaeger:


src/main/java/com/salesforce/dynamodbv2/mt/mappers/sharedtable/impl/BinaryFieldPrefixFunction.java
```diff
@@ -49,7 +49,7 @@ public ByteBuffer apply(FieldValue<ByteBuffer> fieldValue) {
         // equivalent of String.startsWith
         final ByteBuffer prefix = newBuffer(context, tableName, 0).flip();
         final int end = prefix.limit();
-        return b -> b.asReadOnlyBuffer().mismatch(prefix.asReadOnlyBuffer()) == end;
+        return b -> b.mismatch(prefix) == end;
```
busjaeger: I realized it's not necessary to create read views, because `mismatch` doesn't actually advance the buffer position.

PR #456, opened 2019-07-11
misc cleanup

horwitzj:


src/main/java/com/salesforce/dynamodbv2/mt/mappers/sharedtable/impl/MtAmazonDynamoDbBySharedTable.java
```diff
@@ -80,7 +80,6 @@
 
     private final String name;
 
-    private final MeterRegistry meterRegistry;
```
horwitzj: it appears that this field was never read—a `MeterRegistry` value comes into the constructor, but it's only use is when it's passed to `super` (in the constructor)

PR #461, opened 2019-07-12
Introduce rate limiter into streams cache

busjaeger:
Wanted to solicit feedback on this change. (I have not written tests for it yet, so not ready to be merged.)

### DIFF
--- a/src/main/java/com/salesforce/dynamodbv2/mt/util/CachingAmazonDynamoDbStreams.java
+++ b/src/main/java/com/salesforce/dynamodbv2/mt/util/CachingAmazonDynamoDbStreams.java
@@ -8,6 +8,7 @@ import static com.google.common.base.Preconditions.checkArgument;
 import static com.google.common.base.Preconditions.checkNotNull;
 import static com.google.common.collect.Iterables.getLast;
 import static com.salesforce.dynamodbv2.mt.util.ShardIterator.ITERATOR_SEPARATOR;
+import static java.util.Collections.emptyList;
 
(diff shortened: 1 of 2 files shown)


src/main/java/com/salesforce/dynamodbv2/mt/util/CachingAmazonDynamoDbStreams.java
```diff
@@ -773,6 +832,20 @@ public GetRecordsResult getRecords(GetRecordsRequest request) {
         });
     }
 
+    private boolean acquirePermit(StreamShardPosition position) {
+        final boolean[] acquired = new boolean[] { false };
+        try {
+            permitCache.get(position, () -> {
```
msgroi: I guess this is here so as to avoid hitting the rate limiter unnecessarily?

busjaeger: The cache itself itself is the rate limiter. The cache is configured to evict each entry after a certain period of time (by default 100ms). If `get` calls callable, it means there is no entry or the current entry is expired. If it doesn't call the callable, then there is a non-expired entry.

busjaeger:
I thought more about this approach and I think it can have undesired consequences. What I am trying to guard against is multiple threads asking the stream for records at head and get empty results. I believe this approach prevents that, but also has the undesirable side effect that if there are records, one of the threads won't return any. I am working on an alternative solution.

PR #463, opened 2019-07-13
Cache empty results and reduce overlapping reads

busjaeger:
@msgroi  Here is another attempt at fixing the same issue that I think will work better.


busjaeger:
Ready for review.

Based on our metrics (https://www.hostedgraphite.com/c724e2f6/grafana/d/IMSKRfIWk/streams?orgId=2), these changes should help reduce latency and increase throughput.

PR #467, opened 2019-07-16
Cache mapping from TRIM_HORIZON to first loaded sequence number

busjaeger:


src/main/java/com/salesforce/dynamodbv2/mt/util/CachingAmazonDynamoDbStreams.java
```diff
@@ -783,6 +827,15 @@ public GetShardIteratorResult getShardIterator(GetShardIteratorRequest request)
...
             case TRIM_HORIZON:
+                CachingShardIterator cachedIterator =
+                    trimHorizonCache.getIfPresent(new StreamShardId(request.getStreamArn(), request.getShardId()));
+                if (cachedIterator != null) {
+                    return new GetShardIteratorResult().withShardIterator(cachedIterator.toExternalString());
+                } else {
+                    dynamoDbIterator = loadShardIterator(request);
+                    getShardIteratorUncached.increment();
+                }
+                break;
```
busjaeger: Instead of always looking up a physical iterator for trim horizon, it now tries to return a virtual one if the sequence number of the first record in the stream is know. There is a bit of a risk of handing this iterator back and then having trim horizon move forward. So if the iterator is used later and we no longer have those trimmed records in the cache, the client could get a trim horizon exception. I think KCL retries in that case though.

msgroi: Maybe add this as a comment in some form.

src/test/java/com/salesforce/dynamodbv2/mt/util/CachingAmazonDynamoDbStreamsTest.java
```diff
@@ -545,20 +543,71 @@ void testLatest() {
...
+        final String iterator2 = cachingStreams.getShardIterator(iteratorRequest).getShardIterator();
+
+        // first thread looks up records: no mapping cached, so results in call to DynamoDB
+        assertGetRecords(cachingStreams, iterator, null, records);
+        // second thread looks up records: mapping should be cached now, so even though first thread obtained phsyical
+        // iterator, its mapped to cached records without call to DynamoDB (limit so it doesn't try to look beyond)
+        assertGetRecords(cachingStreams, iterator2, 10, records);
+
+        assertCacheMisses(streams, 2, 1);
+    }
```
busjaeger: @msgroi FYI, added another test case that was missing.

msgroi: 👍

src/main/java/com/salesforce/dynamodbv2/mt/util/CachingAmazonDynamoDbStreams.java
busjaeger: Done https://github.com/salesforce/mt-dynamo/pull/467/commits/2e5b888851be86764fcc27db29d25a5ad81a254d

PR #469, opened 2019-07-16
Fix dependency:analyze violations

busjaeger:


pom.xml
```diff
@@ -75,6 +75,11 @@
             <artifactId>micrometer-core</artifactId>
             <version>1.2.0</version>
         </dependency>
+        <dependency>
+            <groupId>org.jetbrains</groupId>
+            <artifactId>annotations</artifactId>
+            <version>13.0</version>
```
horwitzj: 17.0.0?

horwitzj: (https://mvnrepository.com/artifact/org.jetbrains/annotations/17.0.0)

busjaeger: That's what `dependency:analyze` asked for. Let me try with 17.0.0

busjaeger: OK, seems to be happy with that as well.

Issue #462, opened 2019-07-12
Javadoc HTML generation disabled


sbabu-salesforce:
By introducing Kotlin into mt-dynamo, we lost support for generating javadocs. 

issue 1: `mvn javadoc` not supported with kotlin (https://kotlinlang.org/docs/reference/using-maven.html#generating-documentation)
issue 2: Workaround for issue 1 (`dokka`) is not compatible with jdk11 (https://github.com/Kotlin/dokka/issues/294)

We could in theory run dokka running jdk8, but that involves two different build jobs, one for actually building the repo, and the other for generating javadocs. Additionally, that breaks if we start depending on jdk11 exclusive resources, as the jdk8 javadoc generation will not understand said resources.

PR #470, opened 2019-07-16
Disable javadoc to fix release:perform failure

busjaeger:


msgroi:
This is actually related to https://github.com/salesforce/mt-dynamo/issues/462

PR #472, opened 2019-07-16
add better support for pagination matching dynamo spec and test case

sbabu-salesforce:


src/test/java/com/salesforce/dynamodbv2/mt/mappers/MtAmazonDynamoDbTest.java
```diff
@@ -18,23 +22,33 @@
...
+        Set<String> allTableSet = new HashSet<>();
+        ListTablesResult listTablesResult;
+        String exclusiveStartKey = null;
+        int lastBatchSize;
+        int limit = 5;
+        do {
+            listTablesResult = testArgument.getAmazonDynamoDb().listTables(exclusiveStartKey, limit);
+            lastBatchSize = listTablesResult.getTableNames().size();
+            assertTrue(listTablesResult.getTableNames().size() > 0 || lastBatchSize == limit,
+                "Should not have gotten empty table set, or last result set should have been full");
```
sbabu-salesforce: this assertion should address the bug BenB found of always populating last evaluated table name, which causes a client to need to pull an extra empty batch at the end. now that only happens if the previous batch is full, which matches dynamos behavior

src/main/java/com/salesforce/dynamodbv2/mt/mappers/MtAmazonDynamoDbBase.java
```diff
@@ -364,9 +364,16 @@ public ListTablesResult listTables(String exclusiveStartTableName, Integer limit
...
 
-            return new ListTablesResult()
-                .withLastEvaluatedTableName(tableNames.isEmpty() ? null : tableNames.get(tableNames.size() - 1))
-                .withTableNames(tableNames.size() > limit ? tableNames.subList(0, limit) : tableNames);
+            // if there's more results left to go through, populate the lastEvaluatedTableName so the client can
+            // page through for more results.
+            if (tableNames.size() > limit || innerExclusiveStartTableName != null) {
+                return new ListTablesResult()
+                    .withLastEvaluatedTableName(tableNames.get(Math.min(tableNames.size() - 1, limit - 1)))
+                    .withTableNames(tableNames.size() > limit ? tableNames.subList(0, limit) : tableNames);
```
busjaeger: I feel like introducing a local variable could make this code a little simpler:
```
final List<String> tableNames = tableNames.size() > limit ? tableNames.subList(0, limit) : tableNames;
return new ListTablesResult().withLastEvaluatedTableName(Iterables.getLast(tableNames)).withTableNames(tableNames)
```

sbabu-salesforce: that snippet doesn't hold the contract of only setting the last evaluated name if the result set is not full, unfortunately, as you pointed out in a previous review. If we always set lastEvaluatedTableName on the last non-full page of results, we're requiring the clients to pull one more page of empty results from the actual last table name.

sbabu-salesforce: oh you mean within this if block when we know we have to set the last table name. Makes sense, will update

sbabu-salesforce: that said, `Iterables.getLast(tableNames)` is insufficient if `tableNames.size() > limit`, in which case we'd drop the items between `limit` and `tableNames.size()` to the ground. Will keep that snippet as is (`.withLastEvaluatedTableName(tableNames.get(Math.min(tableNames.size() - 1, limit - 1)))`)

PR #473, opened 2019-07-16
relax javadoc requirements

sbabu-salesforce:


src/test/resources/checkstyle.xml
```diff
@@ -235,11 +235,6 @@ Change log:
         </module>
         <module name="NonEmptyAtclauseDescription"/>
         <module name="JavadocTagContinuationIndentation"/>
-        <module name="SummaryJavadoc">
-            <property name="forbiddenSummaryFragments"
-              value="^@return the *|^This method returns |^A [{]@code [a-zA-Z0-9]+[}]( is a )"/>
-        </module>
-        <module name="JavadocParagraph"/>
```
sbabu-salesforce: reiterating comment from previous PR: 

I don't agree with requiring prefixing paragraphs in javadocs with <p> as adding value, instead, I think tedious requirements like that actually dissuade terse javadoc additions, and instead, leads to writing functions with empty javadocs, to "document in a later PR", which also means "never"

src/test/resources/checkstyle.xml
```diff
@@ -235,11 +235,6 @@ Change log:
         </module>
         <module name="NonEmptyAtclauseDescription"/>
         <module name="JavadocTagContinuationIndentation"/>
-        <module name="SummaryJavadoc">
```
sbabu-salesforce: I find for many methods, this leads to redundant code. For basic functions, a javadoc
```
@returns an object transformed from input with blah blah behavior
```
seems totally sufficient. By requiring both a description and a returns flag, there is often comment duplication, resulting in methods I can point out in this repo like: 
```
        /**
         * Returns a new virtual shard iterator positioned at the sequence number of the first record in the list.
         *
         * @param records Record list. Must not be empty.
         * @return New shard iterator positioned at the first record in the list.
         */
        CachingShardIterator resolvePositionIterator(List<Record> records) {
```
```
        /**
         * Returns the DynamoDB shard iterator, may be empty.
         *
         * @return DynamoDB shard iterator
         */
        Optional<String> getDynamoDbIterator() {
```

```
        /**
         * Returns a result object for the given records that starts with a next iterator positioned after the last
         * record.
         *
         * @param records Record list. Must not be empty.
         * @return GetRecordsResult.
         */
```

In each of the above case, either the summary or the returns snippet can be condensed into a single snippet. I'm all about good documentation, but also, I believe in keeping it concise and to the point. More words in javadocs doesn't mean more value. 

```
 For functions doing interesting things, sure add both a description, and param/return snippets, but don't enforce it an all methods is my opinion.

horwitzj: those examples offer something for everyone—those who prefer to look at the summary, those who prefer to look at `@param`/`@return`, and those who love 'em both!

if a developer feels that both are redundant, couldn't they leave out the `@return` part? I certainly do that from time to time.

Google (https://google.github.io/styleguide/javaguide.html#s7.2-summary-fragment) and Oracle (https://www.oracle.com/technetwork/java/javase/documentation/index-137868.html#firstsentence) both note that that each Javadoc block begins with summary text—Google even explicitly notes that Javadoc that has only `@return` code should be replaced with a summary.

src/test/resources/checkstyle.xml
horwitzj: without `<p>` to separate paragraphs, text will appear as one block (except when directly looking at the code)—e.g., generated HTML, IDE mouseover, etc.

the value of offering uniform, readable, and standardized documentation in a public-facing (open-source) project is hard to overestimate.

"document in a later PR" is actually an excellent task for someone learning about the project—whether someone at Salesforce or someone outside the company. (the ones currently in the code were mostly there for Javadocs that were missing when we started using Checkstyle.)

these requirements definitely feel tedious at first (several ZOS folks (including me!) have noted the same)—I hope that fades for you quickly.

PR #475, opened 2019-07-17
add skip javadoc generation flag to release plugin

sbabu-salesforce:
I didn't actually manage to repro the failure in trying to release a new version of mt-dynamo, this change is motivated by this thread: 
https://stackoverflow.com/questions/7412016/how-can-i-disable-the-maven-javadoc-plugin-from-the-command-line

If you read the 2nd answer, one of the strategies proposed is the same BenB used to release the latest mt-dynamo (`release:prepare release:perform -Darguments="-Dmaven.javadoc.skip=true"`). A comment to that answer raises a pom.xml change that achieves the same, which is this PR. 

My approach to repro the failure was fork latest mt-dynamo to my personal repo, modify the git source configs for the plugin to my personal repo, and rename the artifact to `mt-dynamo-temp` (see https://github.com/sbabu-salesforce/mt-dynamo/commit/1e1d21501fa69589ab84690f9e453f84ae227746). 

Unfortunately, that approach failed with a 401 from nexus:
```
[INFO] [ERROR] Failed to execute goal org.apache.maven.plugins:maven-deploy-plugin:2.7:deploy (default-deploy) on project mt-dynamodb-temp: Failed to deploy artifacts: Could not transfer artifact com.salesforce.dynamodb:mt-dynamodb-temp:jar:0.9.47 from/to nexus-releases (https://nexus.soma.salesforce.com/nexus/content/repositories/releases/): Failed to transfer file https://nexus.soma.salesforce.com/nexus/content/repositories/releases/com/salesforce/dynamodb/mt-dynamodb-temp/0.9.47/mt-dynamodb-temp-0.9.47.jar with status code 401 -> [Help 1]
```

I'm guessing I don't have access to release to nexus to that artifact name, or maybe something about the -temp artifact not existing. Happy to test this out with some advice, I'm hesitant to try pushing a release to mt-dymamo having never done so before.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -273,6 +273,9 @@
                 <groupId>org.apache.maven.plugins</groupId>
                 <artifactId>maven-release-plugin</artifactId>
                 <version>2.5.3</version>
+                <configuration>
+                    <arguments>-Dmaven.javadoc.skip=true</arguments>
+                </configuration>
             </plugin>
             <plugin>
                 <artifactId>maven-checkstyle-plugin</artifactId>


msgroi:
The `401` is likely caused by the fact that you don't have permission to write to that nexus folder.  If you wanted that to work you need to open a Concierge ticket.  The https://git.soma.salesforce.com/salesforce-zero/mt-dynamo/blob/master/README.md explains this.  Anyways, that's not necessary.  I suspect that if you've reached the point in the release:perform where it's pushing to nexus you have already passed the javadoc step.  I would think you would see that in the log just above the failed push to nexus.  In summary, I think you've tested this sufficiently.  However, Ben should comment since I thought he had tried this particular workaround and it didn't work.

This would all go away if we made mt-dynamo a git subrepo of zos, which we really should do.  It would pay for itself fairly quickly given how many headaches the maven release plugin and nexus cause.

msgroi:
Related to https://github.com/salesforce/mt-dynamo/issues/462

PR #476, opened 2019-07-17
clean up scanTenantKey overriding and convert TenantTableMetadata to …

sbabu-salesforce:
…MtCreateTableRequest

### DIFF
--- a/src/main/java/com/salesforce/dynamodbv2/mt/mappers/MtAmazonDynamoDbBase.java
+++ b/src/main/java/com/salesforce/dynamodbv2/mt/mappers/MtAmazonDynamoDbBase.java
@@ -114,19 +114,25 @@ public class MtAmazonDynamoDbBase implements MtAmazonDynamoDb {
      * Special default "column" key returned to client on multitenant scans.
      * Configurable by clients if needed.
      */
-    public static final String DEFAULT_SCAN_TENANT_KEY = "mt:context";
-    public static final String DEFAULT_SCAN_VIRTUAL_TABLE_KEY = "mt:tableName";
+    private static final String DEFAULT_SCAN_TENANT_KEY = "mt:context";
+    private static final String DEFAULT_SCAN_VIRTUAL_TABLE_KEY = "mt:tableName";
 
     private final MtAmazonDynamoDbContextProvider mtContext;
     private final AmazonDynamoDB amazonDynamoDb;
     private final MeterRegistry meterRegistry;
+    protected final String scanTenantKey;
+    protected final String scanVirtualTableKey;
 
     protected MtAmazonDynamoDbBase(MtAmazonDynamoDbContextProvider mtContext,
                                    AmazonDynamoDB amazonDynamoDb,
-                                   MeterRegistry meterRegistry) {
+                                   MeterRegistry meterRegistry,
+                                   String scanTenantKey,
+                                   String scanVirtualTableKey) {
         this.mtContext = mtContext;
         this.amazonDynamoDb = amazonDynamoDb;
         this.meterRegistry = meterRegistry;
+        this.scanVirtualTableKey = scanVirtualTableKey == null ? DEFAULT_SCAN_VIRTUAL_TABLE_KEY : scanVirtualTableKey;
+        this.scanTenantKey = scanTenantKey == null ? DEFAULT_SCAN_TENANT_KEY : scanTenantKey;
     }
 
     public AmazonDynamoDB getAmazonDynamoDb() {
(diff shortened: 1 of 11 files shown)


src/main/java/com/salesforce/dynamodbv2/mt/mappers/MtAmazonDynamoDbLogger.java
```diff
@@ -64,7 +64,7 @@ private MtAmazonDynamoDbLogger(MtAmazonDynamoDbContextProvider mtContext,
                                    Consumer<List<String>> logCallback,
                                    List<String> methodsToLog,
                                    boolean logAll) {
-        super(mtContext, amazonDynamoDb, null);
+        super(mtContext, amazonDynamoDb, null, null, null);
```
sbabu-salesforce: I figured forcing instances to null these values consciously is better than overloading a constructor default nulling this. Open to doing that overload given the defaults works for 99.9% of instances I bet

src/test/java/com/salesforce/dynamodbv2/mt/repo/MtDynamoDbTableDescriptionRepoTest.java
```diff
@@ -121,8 +121,41 @@ void testMtDynamoDbTableDescriptionPayPerRequestIsSet() throws InterruptedExcept
...
+    void testListVirtualTables() {
         MtDynamoDbTableDescriptionRepo repo = mtDynamoDbTableDescriptionRepoBuilder.build();
+        List<MtCreateTableRequest> tablesCreated = createPairOfVirtualTables(repo);
+        ListMetadataResult returnedMetadatas =
+            ((MtTableDescriptionRepo)repo).listVirtualTableMetadata(new ListMetadataRequest());
+        assertEquals(new ListMetadataResult(tablesCreated, null), returnedMetadatas);
+    }
+
+    @Test
+    void testListVirtualTables_pagination() {
```
sbabu-salesforce: its worth mentioning by splitting this out into 3 test cases vs 1, I added 10 seconds to the test suite. Each test case creates its own description table with a unique table name including the timestamp since epoch, which translates to at least 5 seconds per test case. 

Our test runs are already pretty slow, this is sort of a bummer, wondering how we should try conserving there.

src/main/java/com/salesforce/dynamodbv2/mt/repo/MtDynamoDbTableDescriptionRepo.java
```diff
@@ -161,22 +171,14 @@ private TenantTable getTenantTableFromHashKey(String hashKey) {
...
-        try {
-            cache.get(tableDescriptionTableName, () -> {
-                createTableDescriptionTableIfNotExists(pollIntervalSeconds);
-                return new TableDescription().withTableName(tableDescriptionTableName);
-            });
-        } catch (ExecutionException e) {
-            throw new RuntimeException(e);
-        }
-        return tableDescriptionTableName;
+        return tableDescriptionSupplier.get().getTableName();
```
sbabu-salesforce: brought this up in slack last week, I found it weird that we were caching this global table name per tenant context, I think to lazy create the metadata table using an existing tried and tested pattern. Ran into this on the empty virtual list table test case, given the scan tried running against a non-existant table. rather than forcing the table creation for that test, I tried cleaning this up as promised.

src/main/java/com/salesforce/dynamodbv2/mt/mappers/MtAmazonDynamoDbBase.java
```diff
@@ -113,19 +113,25 @@
     /**
      * Special default "column" key returned to client on multitenant scans. Configurable by clients if needed.
      */
-    public static final String DEFAULT_SCAN_TENANT_KEY = "mt:context";
-    public static final String DEFAULT_SCAN_VIRTUAL_TABLE_KEY = "mt:tableName";
+    private static final String DEFAULT_SCAN_TENANT_KEY = "mt:context";
+    private static final String DEFAULT_SCAN_VIRTUAL_TABLE_KEY = "mt:tableName";
```
busjaeger: It looks like in other cases defaults and defaulting behavior is in the builder. To stay consistent with that, it may make sense to do the same here.

src/main/java/com/salesforce/dynamodbv2/mt/repo/MtDynamoDbTableDescriptionRepo.java
```diff
@@ -304,12 +302,13 @@ public ListMetadataResult listVirtualTableMetadata(ListMetadataRequest listMetad
         scanReq.setExclusiveStartKey(lastEvaluatedKey);
         scanReq.setLimit(listMetadataRequest.getLimit());
         scanResult = amazonDynamoDb.scan(scanReq);
-        List<TenantTableMetadata> metadataList = scanResult.getItems().stream()
+        List<MtCreateTableRequest> metadataList = scanResult.getItems().stream()
```
horwitzj: should this variable name be updated as well?

PR #477, opened 2019-07-17
Initial backup restore

sbabu-salesforce:


sbabu-salesforce:
Made some more progress fleshing out the backup restore pipeline, and removing changes not directly related to backup-restores. 

I'd start reviewing `MtAmazonDynamoDbBySharedTable` to see how clients would call backups and restores, then checkout `MtSharedTableBackupManagerImplTest` to see a basic backup-restore run successfully, and finally go through the backup restore code (`MtBackupManager` and `MtSharedTableBackupManagerImpl`)

src/test/kotlin/com/salesforce/dynamodbv2/mt/mappers/sharedtable/impl/MtSharedTableBackupManagerImplTest.kt
```diff
@@ -0,0 +1,160 @@
...
+                .build()
+
+        val backupManager = sharedTableBinaryHashKey.backupManager
+
+        var s3: AmazonS3? = null
+
+        @BeforeAll
+        @JvmStatic
+        internal fun beforeAll() {
+            // TODO: Make this test operate against mock s3 in an autogenerated bucket
```
sbabu-salesforce: grrr, I did spend a good amount of time trying to use a mock s3 implementation that zero-importer uses in its unit tests, but ran into maven dependency hell. The local dynamo db jar pulls in dependencies for a super old jetty server jar, which collides with the mock s3 newer jetty server jar, and blows up. Tried a few things unsuccessfully, and got frustrated. 

Debating switching the unit tests over to operate against dynamo in docker, so the JVM doesn't have to pull in ancient jetty jars, but that's non-ideal given the number of test cases we have using local dynamo and how many local flows would get disrupted. 

I guess I can go the other way, and setup a mock s3 in a local docker like Brian has in the local universe, and have our test suite do a docker compose and spin up a local s3? Thoughts?

sbabu-salesforce:
It's worth mentioned my planned next steps are to get through, at which point I hope this is good to merge (but shouldn't stop reviews from starting). Once this merges, I plan to start tackling the ZOS side of things to create a worker dyno actually calling this backup code (on demand, eventually periodically), and trying out real restores of real ATSIS perf/staging data. 
- implement describeBackup and listBackups for ZOS to consume
- operate against quiesced table (on-demand backup/restored data and metadata tables)
- get this working against mock s3
- flush out client (ZOS) facing backup request/response objects, and restore request/response objects

Things I think we can defer to after seeing this on ZOS (or during, but after this merge)
- Add resume logic, so backups can survive a dyno crash while generating the backup (properly async). For now, I think marking a backup left in progress as failed is fine, to get this going and tested.
- Add support for backing up byTable and byAccount (how important is this?)
- Improve serialization logic from json
- Move backup metadata from S3 files to Dynamo (structured metadata will make management miles easier down the line, and I'm worried it'll get annoying to manage medium-long term)

src/main/java/com/salesforce/dynamodbv2/mt/mappers/sharedtable/impl/MtAmazonDynamoDbBySharedTable.java
```diff
@@ -551,6 +590,119 @@ public UpdateItemResult updateItem(UpdateItemRequest updateItemRequest) {
...
+            throw new ContinuousBackupsUnavailableException("Backups can only be created by configuring a backup "
+                + "managed on an mt-dynamo table builder, see <insert link to backup guide>");
+        }
+    }
+
+    @Override
+    public RestoreTableFromBackupResult restoreTableFromBackup(
+        RestoreTableFromBackupRequest restoreTableFromBackupRequest) {
+        if (backupManager.isPresent()) {
+            if (!(restoreTableFromBackupRequest instanceof RestoreMtBackupRequest)) {
```
busjaeger: This is the first case where we require a subtype of a request object be passed to the API. Would be good to review whether there are possible alternatives. For example, could we encode whatever additional data we need in the backupArn?

src/main/java/com/salesforce/dynamodbv2/mt/mappers/sharedtable/impl/MtAmazonDynamoDbBySharedTable.java
```diff
@@ -551,6 +591,121 @@ public UpdateItemResult updateItem(UpdateItemRequest updateItemRequest) {
...
+                + "managed on an mt-dynamo table builder, see <insert link to backup guide>");
+        }
+    }
+
+    @Override
+    public CreateBackupResult createBackup(CreateBackupRequest createBackupRequest) {
+        if (backupManager.isPresent()) {
+            Preconditions.checkNotNull(createBackupRequest.getBackupName(), "Must pass backup name.");
+            Preconditions.checkArgument(createBackupRequest.getTableName() == null,
+                "Multitenant backups cannot backup individual tables, table name arguments are disallowed.");
```
busjaeger: It seems like this should be a separate method that does not implement the `AmazonDynamoDB` interface, since it's creating a backup for all physical tables managed by this shared table instance. Don't feel super strongly about it though, since we can change that later (since ZOS will likely be the only consumer for a while).

src/main/java/com/salesforce/dynamodbv2/mt/mappers/sharedtable/impl/MtAmazonDynamoDbBySharedTable.java
```diff
@@ -551,6 +591,121 @@ public UpdateItemResult updateItem(UpdateItemRequest updateItemRequest) {
...
+                backupManager.get().getMtBackupTableSnapshotter();
+                for (SnapshotResult snapshotResult : snapshotResults) {
+                    backupManager.get().backupPhysicalMtTable(createBackupRequest,
+                        snapshotResult.getTempSnapshotTable());
+                }
+                MtBackupMetadata finishedMetadata = backupManager.get().markBackupComplete(createBackupRequest);
+                return new CreateBackupResult().withBackupDetails(
+                    new BackupDetails()
+                        // TODO: maybe this should be the ARN for the global S3 metadata file for this backup
+                        .withBackupArn(finishedMetadata.getMtBackupName())
```
busjaeger: How will ZOS use this ARN?

src/main/java/com/salesforce/dynamodbv2/mt/mappers/sharedtable/impl/MtAmazonDynamoDbBySharedTable.java
sbabu-salesforce: right, I remember that conversation. Sure, I can move the backup hook to a public method of `MtAmazonDynamoDbBySharedTable` that ZOS can call directly, instead of abusing this API.

src/main/java/com/salesforce/dynamodbv2/mt/mappers/sharedtable/impl/MtAmazonDynamoDbBySharedTable.java
sbabu-salesforce: good q. I'm thinking to return these Arn's in listBackup calls, so tenant context specified calls to listBackup's return backups available to said tenant of all virtual tables (if unspecified) with these ARN's encoded within mt-dynamo, available for restore. 

Will post an update to listBackups accordingly.

src/main/java/com/salesforce/dynamodbv2/mt/mappers/sharedtable/impl/MtAmazonDynamoDbBySharedTable.java
```diff
@@ -551,6 +591,121 @@ public UpdateItemResult updateItem(UpdateItemRequest updateItemRequest) {
...
+        }
+    }
+
+    @Override
+    public RestoreTableFromBackupResult restoreTableFromBackup(
+        RestoreTableFromBackupRequest restoreTableFromBackupRequest) {
+        if (backupManager.isPresent()) {
+
+            TenantTableBackupMetadata backupMetadata = ((MtSharedTableBackupManager) backupManager.get())
+                    .getTenantTableBackupFromArn(restoreTableFromBackupRequest.getBackupArn());
```
busjaeger: Should we verify here that the current thread tenant context matches the backup arn? (Similar to what we do for streams: https://github.com/salesforce/mt-dynamo/blob/master/src/main/java/com/salesforce/dynamodbv2/mt/mappers/MtAmazonDynamoDbStreamsBase.java#L46)

src/main/java/com/salesforce/dynamodbv2/mt/mappers/sharedtable/impl/MtAmazonDynamoDbBySharedTable.java
```diff
@@ -551,6 +591,121 @@ public UpdateItemResult updateItem(UpdateItemRequest updateItemRequest) {
...
+                "Multitenant backups cannot backup individual tables, table name arguments are disallowed.");
+            backupManager.get().createBackup(createBackupRequest);
+
+            ExecutorService executorService = Executors.newFixedThreadPool(mtTables.keySet().size());
+
+            Set<String> snapshottedTables = Sets.newHashSet();
+            List<Future<SnapshotResult>> futures = Lists.newArrayList();
+            Set<String> origMtTables = ImmutableSet.copyOf(mtTables.keySet());
+            for (String tableName : origMtTables) {
+                String snapshottedTable = tableName + "-copy";
```
busjaeger: There are some methods that scan tables to find all shared tables for a given prefix. Should these `-copy` tables be excluded from those scans?

src/main/java/com/salesforce/dynamodbv2/mt/mappers/sharedtable/impl/MtAmazonDynamoDbBySharedTable.java
```diff
@@ -551,6 +591,121 @@ public UpdateItemResult updateItem(UpdateItemRequest updateItemRequest) {
...
+                    } catch (InterruptedException | ExecutionException e) {
+                        throw new MtBackupException("Error snapshotting table", e);
+                    }
+                }
+            } finally {
+                executorService.shutdown();
+            }
+            try {
+                backupManager.get().getMtBackupTableSnapshotter();
+                for (SnapshotResult snapshotResult : snapshotResults) {
```
busjaeger: Why not do this in parallel as well?

src/main/java/com/salesforce/dynamodbv2/mt/mappers/sharedtable/impl/MtAmazonDynamoDbBySharedTable.java
```diff
@@ -551,6 +591,121 @@ public UpdateItemResult updateItem(UpdateItemRequest updateItemRequest) {
...
+                MtBackupMetadata finishedMetadata = backupManager.get().markBackupComplete(createBackupRequest);
+                return new CreateBackupResult().withBackupDetails(
+                    new BackupDetails()
+                        // TODO: maybe this should be the ARN for the global S3 metadata file for this backup
+                        .withBackupArn(finishedMetadata.getMtBackupName())
+                        .withBackupCreationDateTime(new Date(finishedMetadata.getCreationTime()))
+                        .withBackupName(finishedMetadata.getMtBackupName()));
+            } finally {
+                for (SnapshotResult result : snapshotResults) {
+                    backupManager.get().getMtBackupTableSnapshotter().cleanup(result, getAmazonDynamoDb());
```
busjaeger: Should cleanup wait until the `-copy` tables have been successfully deleted? Otherwise, don't we run the risk that the next cycle fails?

src/main/kotlin/com/salesforce/dynamodbv2/mt/backups/MtBackupAwsAdaptor.kt
```diff
@@ -0,0 +1,46 @@
...
+import com.amazonaws.services.dynamodbv2.model.BackupDetails
+import com.amazonaws.services.dynamodbv2.model.BackupStatus
+import com.amazonaws.services.dynamodbv2.model.BackupSummary
+import com.amazonaws.services.dynamodbv2.model.DescribeBackupResult
+import java.util.*
+
+/**
+ * Convert simple pojos defined in {@link MtBackupManager} to and from AWS objects defined in AWS specs.
+ */
+class MtBackupAwsAdaptor {
```
busjaeger: Why is this needed?

src/main/kotlin/com/salesforce/dynamodbv2/mt/backups/MtBackupTableSnapshotter.kt
```diff
@@ -0,0 +1,123 @@
...
+        // generate backup of table
+        var backupResult: CreateBackupResult? = null
+        do {
+            try {
+                backupResult = snapshotRequest.amazonDynamoDb.createBackup(CreateBackupRequest()
+                        .withBackupName(snapshotRequest.targetTableName)
+                        .withTableName(snapshotRequest.sourceTableName))
+            } catch (e: ContinuousBackupsUnavailableException) {
+                // wait for backups to be ready on this table, if not already
+                Thread.sleep(1000L)
```
busjaeger: Could use Awaitility https://github.com/salesforce/mt-dynamo/blob/39911ab376ba8dafed59f0c87d4f85d77168d6a0/src/main/java/com/salesforce/dynamodbv2/mt/admin/AmazonDynamoDbAdminUtils.java#L99

src/main/java/com/salesforce/dynamodbv2/mt/mappers/sharedtable/impl/MtAmazonDynamoDbBySharedTable.java
```diff
@@ -551,6 +591,121 @@ public UpdateItemResult updateItem(UpdateItemRequest updateItemRequest) {
...
+            Preconditions.checkNotNull(createBackupRequest.getBackupName(), "Must pass backup name.");
+            Preconditions.checkArgument(createBackupRequest.getTableName() == null,
+                "Multitenant backups cannot backup individual tables, table name arguments are disallowed.");
+            backupManager.get().createBackup(createBackupRequest);
+
+            ExecutorService executorService = Executors.newFixedThreadPool(mtTables.keySet().size());
+
+            Set<String> snapshottedTables = Sets.newHashSet();
+            List<Future<SnapshotResult>> futures = Lists.newArrayList();
+            Set<String> origMtTables = ImmutableSet.copyOf(mtTables.keySet());
```
busjaeger: What about the metadata table? Do we need to take a snapshot of that one as well?

src/main/kotlin/com/salesforce/dynamodbv2/mt/mappers/sharedtable/impl/MtSharedTableBackupManager.kt
```diff
@@ -0,0 +1,456 @@
...
+                    sharedTableMtDynamo.mtTableDescriptionRepo.listVirtualTableMetadata(
+                            MtTableDescriptionRepo.ListMetadataRequest()
+                                .withExclusiveStartKey(startKey)
+                                .withLimit(batchSize))
+            commitTenantTableMetadata(createBackupRequest.backupName, listTenantMetadataResult)
+            tenantTables.addAll(listTenantMetadataResult.createTableRequests)
+            tenantTableCount += listTenantMetadataResult.createTableRequests.size
+        } while (listTenantMetadataResult.lastEvaluatedTable != null)
+        logger.info("${createBackupRequest.backupName}: Finished generating backup metadata for " +
+                "${tenantTables.size} virtual table in ${System.currentTimeMillis() - startTime} ms")
```
busjaeger: Maybe emit Micrometer metrics for duration?

src/main/kotlin/com/salesforce/dynamodbv2/mt/mappers/sharedtable/impl/MtSharedTableBackupManager.kt
```diff
@@ -0,0 +1,456 @@
...
+        return ret
+    }
+
+    override fun listBackups(listBackupRequest: ListBackupsRequest): ListBackupsResult {
+        val ret = Lists.newArrayList<BackupSummary>()
+
+        Preconditions.checkArgument(listBackupRequest.backupType == null, "Listing backups by backupType unsupported")
+        Preconditions.checkArgument(listBackupRequest.tableName == null, "Listing backups by table name unsupported for multi tenant backups")
+        Preconditions.checkArgument(listBackupRequest.timeRangeLowerBound == null, "Listing backups filtered by time range unsupported [currently]")
+        Preconditions.checkArgument(listBackupRequest.timeRangeUpperBound == null, "Listing backups filtered by time range unsupported [currently]")
```
busjaeger: It would be good to start thinking about how we want to lay out backup metadata in DynamoDB.

src/main/kotlin/com/salesforce/dynamodbv2/mt/mappers/sharedtable/impl/MtSharedTableBackupManager.kt
```diff
@@ -0,0 +1,456 @@
...
+
+    override fun getBackup(id: String): MtBackupMetadata? {
+        try {
+            val backupFile: S3Object = s3.getObject(s3BucketName, getBackupMetadataFile(id))
+            val backupFileString = CharStreams.toString(
+                    InputStreamReader(backupFile.objectContent.delegateStream, charset))
+            return gson.fromJson(backupFileString,
+                    MtBackupMetadata::class.java)
+        } catch (e: AmazonS3Exception) {
+            if ("NoSuchKey".equals(e.errorCode)) {
```
busjaeger: Note: you may get this even if the object exists: https://codeburst.io/quick-explanation-of-the-s3-consistency-model-6c9f325e3f82

src/main/kotlin/com/salesforce/dynamodbv2/mt/mappers/sharedtable/impl/MtSharedTableBackupManager.kt
```diff
@@ -0,0 +1,456 @@
...
+        for (tenantTableMetadata in tenantTableMetadatas.createTableRequests) {
+            val tenantTableMetadataJson = gson.toJson(tenantTableMetadata.createTableRequest).toByteArray(charset)
+            val objectMetadata = ObjectMetadata()
+            objectMetadata.contentLength = tenantTableMetadataJson.size.toLong()
+            objectMetadata.contentType = "application/json"
+            val putObjectReq = PutObjectRequest(s3BucketName,
+                    getTenantTableMetadataFile(backupId, TenantTable(tenantTableMetadata.createTableRequest.tableName, tenantTableMetadata.tenantName)),
+                    gson.toJson(tenantTableMetadata.createTableRequest).byteInputStream(charset),
+                    objectMetadata)
+            s3.putObject(putObjectReq)
```
busjaeger: Note: reads may not return the latest metadata after updating it: https://codeburst.io/quick-explanation-of-the-s3-consistency-model-6c9f325e3f82

src/main/kotlin/com/salesforce/dynamodbv2/mt/mappers/sharedtable/impl/MtSharedTableBackupManager.kt
```diff
@@ -0,0 +1,456 @@
...
+                    .linkedHashKeys().hashSetValues().build<TenantTable, TenantTableRow>()
+            for (i in 0..(scanResult.items.size - 1)) {
+                val row = scanResult.items[i]
+                val tenant = row.get(sharedTableMtDynamo.scanTenantKey)!!.s
+                val virtualTable = row.get(sharedTableMtDynamo.scanVirtualTableKey)!!.s
+                rowsPerTenant.put(TenantTable(tenantName = tenant, virtualTableName = virtualTable),
+                        TenantTableRow(scanResult.items[i]))
+            }
+
+            flushScanResult(createBackupRequest.backupName, numPasses, rowsPerTenant)
```
busjaeger: Do we run the risk of creating many small s3 files, given the current shared table layout (tenant id & table name in hash key)?

busjaeger:
You may also want to consider squashing some of the commits to keep the history simple.

src/main/kotlin/com/salesforce/dynamodbv2/mt/backups/MtBackupTableSnapshotter.kt
sbabu-salesforce: makes sense, will include that in followup PR

src/main/kotlin/com/salesforce/dynamodbv2/mt/backups/MtBackupAwsAdaptor.kt
sbabu-salesforce: not strictly necessary, sort of a utility class for easy unit testing

src/main/java/com/salesforce/dynamodbv2/mt/mappers/sharedtable/impl/MtAmazonDynamoDbBySharedTable.java
sbabu-salesforce: makes sense, will include that in follow up PR

src/main/java/com/salesforce/dynamodbv2/mt/mappers/sharedtable/impl/MtAmazonDynamoDbBySharedTable.java
sbabu-salesforce: makes sense, will include in follow up

src/main/java/com/salesforce/dynamodbv2/mt/mappers/sharedtable/impl/MtAmazonDynamoDbBySharedTable.java
sbabu-salesforce: Given we don't allow updates to existing virtual table metadata, I don't think we do need a snapshot. A row either appears, or does not, and for both cases, missing metadata for virtual data should be handled as an orphaned backup, and vice versa as an empty backup

src/main/java/com/salesforce/dynamodbv2/mt/mappers/sharedtable/impl/MtAmazonDynamoDbBySharedTable.java
sbabu-salesforce: will address that in follow up PR

src/main/kotlin/com/salesforce/dynamodbv2/mt/mappers/sharedtable/impl/MtSharedTableBackupManager.kt
sbabu-salesforce: yes, that's a risk. I figure we can optimize that if/when it becomes a problem.

src/main/kotlin/com/salesforce/dynamodbv2/mt/mappers/sharedtable/impl/MtSharedTableBackupManager.kt
sbabu-salesforce: sure, will keep that close to the top of the next steps.

src/main/kotlin/com/salesforce/dynamodbv2/mt/mappers/sharedtable/impl/MtSharedTableBackupManager.kt
sbabu-salesforce: I think we can live with that risk till we move metadata to dynamo, I mostly just want to see some real backups and restores in action before spending time on that

src/main/kotlin/com/salesforce/dynamodbv2/mt/mappers/sharedtable/impl/MtSharedTableBackupManager.kt
```diff
@@ -0,0 +1,456 @@
...
+        // restore tenant-table data
+        val backupFileKeys = getBackupFileKeys(restoreMtBackupRequest.backupName,
+                restoreMtBackupRequest.tenantTableBackup)
+        for (fileName in backupFileKeys) {
+            val backupFile: S3Object = s3.getObject(s3BucketName, fileName)
+            val rowsToInsert: List<TenantTableRow> = gson.fromJson(backupFile.objectContent.bufferedReader(),
+                    object : TypeToken<List<TenantTableRow>>() {}.type)
+            for (row in rowsToInsert) {
+                mtContext.withContext(restoreMtBackupRequest.newTenantTable.tenantName) {
+                    sharedTableMtDynamo.putItem(restoreMtBackupRequest.newTenantTable.virtualTableName, row.attributeMap)
```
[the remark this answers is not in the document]
sbabu-salesforce: do we allow adding indexes or removing indexes? If that's not supported yet (which I don't think it is from our conversations), I think the work to support those virtual table updates should include making backups support those changes.

src/main/kotlin/com/salesforce/dynamodbv2/mt/mappers/sharedtable/impl/MtSharedTableBackupManager.kt
sbabu-salesforce: sure, will include in a followup

src/main/java/com/salesforce/dynamodbv2/mt/mappers/sharedtable/impl/MtAmazonDynamoDbBySharedTable.java
sbabu-salesforce: addressed in https://github.com/salesforce/mt-dynamo/pull/524 I think? unless there was something else that needs to filter these temp tables?

src/main/java/com/salesforce/dynamodbv2/mt/mappers/sharedtable/impl/MtAmazonDynamoDbBySharedTable.java
sbabu-salesforce: addressed in https://github.com/salesforce/mt-dynamo/pull/525

src/main/java/com/salesforce/dynamodbv2/mt/mappers/sharedtable/impl/MtAmazonDynamoDbBySharedTable.java
sbabu-salesforce: addressed in https://github.com/salesforce/mt-dynamo/pull/525

PR #486, opened 2019-07-24
Make CompositeStrings jdk8 compliant

brian-mcnamara:


src/main/java/com/salesforce/dynamodbv2/mt/util/CompositeStrings.java
```diff
@@ -64,7 +64,7 @@ public String join(Iterable<String> values) {
      * @return Sequence of keys contained in the given composite key
      */
     public Iterator<String> split(String composite) {
-        return new Iterator<>() {
+        return new Iterator<String>() {
```
busjaeger: Is mt-dynamo going to stay in jdk11 mode? Are we running the risk that we accidentally introduce issues like this again without catching it as part of the mt-dynamo build?

brian-mcnamara:
Closing, after talking deciding to just copy the class to keep things simple on the other end.

PR #489, opened 2019-07-25
Record age of cache entries evicted from streams records cache

busjaeger:


src/main/java/com/salesforce/dynamodbv2/mt/util/StreamsRecordCache.java
```diff
@@ -426,6 +439,7 @@ private void evictRecords() {
                                 numEvicted += evicted.getRecords().size();
                                 size.addAndGet(-evicted.getRecords().size());
                                 byteSize.addAndGet(-evicted.getByteSize());
+                                evictRecordsAge.record(System.nanoTime() - evicted.getCreateTime(), NANOSECONDS);
```
sbabu-salesforce: This is tracking the amount of time a set of records lives in the cache before being evicted ya? Makes sense. For my own information, which metric tracks the amount of time a record appears in the stream before getting populated into the cache?

I see us updating metrics with the size of records inserted, but not the age of the records when populated

busjaeger: Good question :). I debated which metrics to add. Each DynamoDB stream record has a creation timestamp (`StreamRecord#getApproximateCreationDateTime`). So originally, I was thinking of having two timers, one that records age of records added and one for age of records evicted.

I went with this approach, because I am mainly after a metric to assess the size of our cache, i.e., I figured short time in the cache would be a good indicator that our cache is too small.

Having said that, I feel like the other timers would also give us valuable info. Not sure if they would make the time-in-cache timer obsolete.

Thoughts?

sbabu-salesforce: I think it might be something useful to track, it'd be the first thing to look at if a customer claims something on the delta API showed up later (or never) when investigating, but no biggie, just a curiosity of mine. It should be easy enough to add in if/when we need to

busjaeger: Makes sense. FWIW, we do track returned record age in the ZOS delta API (see topmost chart here: https://www.hostedgraphite.com/c724e2f6/grafana/d/IMSKRfIWk/streams?orgId=2&from=now-3h&to=now&refresh=5s&var-proctype=zos-streams&var-appname=sfdc-0-objects-ci)

PR #497, opened 2019-07-30
bump ktlint version

horwitzj:


horwitzj:
it looks like
```fun formatScript(
        text: String,
        ruleSets: Iterable<RuleSet>,
        userData: Map<String, String>,
        cb: (e: LintError, corrected: Boolean) -> Unit
    )```
was removed from ktlint (but spotless seems to want it to be there): https://github.com/pinterest/ktlint/commit/ba513d7b164e9e98deafec68b2888101d8c34165

pallavkothari:
tracked here - https://github.com/diffplug/spotless/issues/419

PR #500, opened 2019-08-05
simplify Optional.isEmpty() assertions

horwitzj:


sbabu-salesforce:
I should mention, one advantage of `assertEquals(Optional.empty(), optionalValue)` over `assertTrue(optionalValue.isEmpty())`, is, when optionalValue is not empty, we get no error message about what its populated to, where the former gives us the value in the assertion failure message. To keep that behavior, I'd either leave the code as is, or do: `assertTrue(optionalValue.isEmpty(), optionalValue.toString())`

It's a corner case, but when the test fails there, its nice to have that info

horwitzj:
this is an excellent point and an excellent suggestion

PR #506, opened 2019-08-08
fix for issue#505

msgroi:
fix for issue#505

### DIFF
--- a/src/main/java/com/salesforce/dynamodbv2/mt/repo/MtDynamoDbTableDescriptionRepo.java
+++ b/src/main/java/com/salesforce/dynamodbv2/mt/repo/MtDynamoDbTableDescriptionRepo.java
@@ -106,9 +106,9 @@ public class MtDynamoDbTableDescriptionRepo implements MtTableDescriptionRepo {
 
     @Override
     public TableDescription createTable(CreateTableRequest createTableRequest) {
-        amazonDynamoDb.putItem(new PutItemRequest().withTableName(getTableDescriptionTableName())
-            .withItem(createItem(createTableRequest)));
-        return getTableDescription(createTableRequest.getTableName());
+        Map<String, AttributeValue> item = createItem(createTableRequest);
+        amazonDynamoDb.putItem(new PutItemRequest().withTableName(getTableDescriptionTableName()).withItem(item));
+        return itemToDescription(item);
     }
 
     @Override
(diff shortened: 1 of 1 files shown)


src/main/java/com/salesforce/dynamodbv2/mt/repo/MtDynamoDbTableDescriptionRepo.java
```diff
@@ -106,9 +106,9 @@ private MtDynamoDbTableDescriptionRepo(AmazonDynamoDB amazonDynamoDb,
 
     @Override
     public TableDescription createTable(CreateTableRequest createTableRequest) {
-        amazonDynamoDb.putItem(new PutItemRequest().withTableName(getTableDescriptionTableName())
-            .withItem(createItem(createTableRequest)));
-        return getTableDescription(createTableRequest.getTableName());
+        Map<String, AttributeValue> item = createItem(createTableRequest);
+        amazonDynamoDb.putItem(new PutItemRequest().withTableName(getTableDescriptionTableName()).withItem(item));
+        return itemToDescription(item);
```
busjaeger: So after the write the description is not placed into the cache? So next time we call describe, we may still not find the table, correct? Would it make sense to switch to strongly consistent reads instead?

msgroi: With this solution the run-time path on the subsequent request against the same host will be ever so slightly slower than it would be if we did the consistent read and loaded up the cache on table create so it seems like hair splitting performance-wise.  The code fix that does the consistent read is more simple but makes every table description load more expensive.  🤷‍♂

busjaeger: I am not so worried about performance, rather the expected behavior of `AmazonDynamoDB#createTable` and `AmazonDynamoDB#describeTable`. If I call the former and get confirmation that the table was successfully created, should I be able to expect that the latter will return its description? Isn't there still a (very small) chance with this implementation that the latter throw a `ResourceNotFoundException`?

msgroi: Yeah, good point.  The likelihood of that is low, but 0% likelihood is better than 0.01% so I fixed it here ... 2a5a7fc6a21b67339e7fc6d569e53a88f22a4735

PR #509, opened 2019-08-10
Fix metrics

degaydorus:
This PR allows for clients to pass in a metric prefix.  It first supported this by allowing cache registration to be done on the client side, using cache builder methods in mt-dynamo.  To maintain encapsulation, this was modified to allow a client to pass in a metric prefix when creating a `CachingAmazonDynamoDbStreams` instance.
- added tests for monitoring caches created in `CachingAmazonDynamoDbStreams`
- added builder for describe stream cache to use in tests (previously exposed getter)

### DIFF
--- a/src/main/java/com/salesforce/dynamodbv2/mt/util/CachingAmazonDynamoDbStreams.java
+++ b/src/main/java/com/salesforce/dynamodbv2/mt/util/CachingAmazonDynamoDbStreams.java
@@ -23,6 +23,7 @@ import com.amazonaws.services.dynamodbv2.model.Record;
 import com.amazonaws.services.dynamodbv2.model.ResourceNotFoundException;
 import com.amazonaws.services.dynamodbv2.model.Shard;
 import com.amazonaws.services.dynamodbv2.model.ShardIteratorType;
+import com.amazonaws.util.StringUtils;
 import com.google.common.annotations.VisibleForTesting;
 import com.google.common.base.Throwables;
 import com.google.common.base.Ticker;
@@ -115,12 +116,10 @@ public class CachingAmazonDynamoDbStreams extends DelegatingAmazonDynamoDbStream
 
         private final AmazonDynamoDBStreams amazonDynamoDbStreams;
         private MeterRegistry meterRegistry;
+        private String metricPrefix;
         private Sleeper sleeper;
         private Ticker ticker;
         private Cache<String, DescribeStreamResult> describeStreamCache;
-        private Cache<StreamShardPosition, Boolean> getRecordsEmptyResultCache;
-        private Cache<CachingShardIterator, String> iteratorCache;
-        private Cache<StreamShardId, CachingShardIterator> trimHorizonCache;
         private long maxRecordsByteSize = DEFAULT_MAX_RECORD_BYTES_CACHED;
         private int maxIteratorCacheSize = DEFAULT_MAX_ITERATOR_CACHE_SIZE;
         private int getRecordsMaxRetries = DEFAULT_GET_RECORDS_MAX_RETRIES;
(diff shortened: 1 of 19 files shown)


src/main/java/com/salesforce/dynamodbv2/mt/util/CachingAmazonDynamoDbStreams.java
```diff
@@ -217,6 +221,17 @@ public Builder withMaxIteratorCacheSize(int maxIteratorCacheSize) {
...
+        /**
+         * Cache to use for describe stream calls.
+         *
+         * @param describeStreamCache The cache to use for describe stream
+         * @return this Builder.
+         */
+        public Builder withDescribeStreamCache(Cache describeStreamCache) {
+            this.describeStreamCache = describeStreamCache;
+            return this;
+        }
```
busjaeger: Do you still need to expose this?

src/main/java/com/salesforce/dynamodbv2/mt/util/CachingAmazonDynamoDbStreams.java
```diff
@@ -720,29 +739,26 @@ private static String toShortString(GetRecordsResult result) {
...
-        GuavaCacheMetrics.monitor(meterRegistry, getRecordsEmptyResultCache, cn + ".EmptyResult");
-        GuavaCacheMetrics.monitor(meterRegistry, describeStreamCache, cn + ".DescribeStream");
-        GuavaCacheMetrics.monitor(meterRegistry, trimHorizonCache, cn + ".TrimHorizon");
-    }
-
-    @VisibleForTesting
-    // TODO: introduce builder/constructor variant that allows passing in cache reference
-    public Cache<String, DescribeStreamResult> getDescribeStreamCache() {
-        return describeStreamCache;
+        final String prefix = StringUtils.isNullOrEmpty(metricPrefix) ? cn : cn + "." + metricPrefix;
```
busjaeger: Just curious: why did you decide to go with this option vs injecting the prefix via a `MeterRegistry` in ZOS?

degaydorus: I wanted to talk to you about that, I did try implementing `MeterRegistry`, and had a couple concerns.  See below are some examples, in creating a timer, I chose to call the method that takes tags, but there are other timer methods (one that takes strings), but there wasn't an option to pass in distributionStatisticConfig for example.  Most of the registries that implement `MeterRegistry` have separate classes for each type, i.e. Composite counter.  I think we just want to add the prefix and not have custom metric types, basically wasn't sure how in depth we wanted to go with the implementation.

    @Override
    protected Timer newTimer(Id id, DistributionStatisticConfig distributionStatisticConfig,
                             PauseDetector pauseDetector) {
        return meterRegistry.timer(prefix + "." + id.getName(), id.getTags());
    }

    @Override
    protected DistributionSummary newDistributionSummary(Id id, DistributionStatisticConfig distributionStatisticConfig,
                                                         double scale) {
        return meterRegistry.summary(id.getName(), id.getTags());
    }

src/main/java/com/salesforce/dynamodbv2/mt/util/CachingAmazonDynamoDbStreams.java
degaydorus: https://github.com/salesforce/mt-dynamo/pull/509/files#diff-2a4b73d2ee26b7ce82d49248a4931568 `MtAmazonDynamoDbStreamsBaseTest ` uses it in the test.  Before there was a getter for the test, in the past we discussed adding a builder instead.  In both options (getter or builder method), the test need a way to get to the cache.

degaydorus:
> As discussed could considering taking out cache builder method.

I removed the builder method for describeStreamCache, added the getter only for `CachingAmazonDynamoDbStreams` because checking the content of what's added in the cache is still necessary for some of those tests.  The getter has access scope reduced to non-modifier instead of public.

`MtAmazonDynamoDbStreamsBaseTest` just checks the meterRegistry because that's all that's needed at that level (not the contents of what's cached).

PR #521, opened 2019-08-20
expose delete backup functionality out for zos to call

sbabu-salesforce:


src/main/java/com/salesforce/dynamodbv2/mt/mappers/sharedtable/impl/MtAmazonDynamoDbBySharedTable.java
```diff
@@ -654,6 +657,21 @@ public DescribeBackupResult describeBackup(DescribeBackupRequest describeBackupR
         }
     }
 
+    @Override
+    public DeleteBackupResult deleteBackup(DeleteBackupRequest deleteBackupRequest) {
```
busjaeger: Is this deleting virtual table or physical table backups?

sbabu-salesforce: just the backup! to clean up files on S3, for when I start running this backup worker dyno, we have a way to clean up behind ourselves.

busjaeger: OK, then I think in the future we may want to consider introducing a separate API for this one as well.

PR #524, opened 2019-08-21
filter out temp copy tables from list table calls

sbabu-salesforce:


src/main/java/com/salesforce/dynamodbv2/mt/mappers/sharedtable/impl/MtAmazonDynamoDbBySharedTable.java
```diff
@@ -182,7 +182,7 @@ Clock getClock() {
 
     @Override
     protected boolean isMtTable(String tableName) {
-        return mtTables.containsKey(tableName);
+        return mtTables.containsKey(tableName) && !tableName.endsWith(BACKUP_TEMP_TABLE_SUFFIX);
```
busjaeger: Does this preclude suffixing a physical table with `-copy`?

sbabu-salesforce: glad you brought that up, we should talk about that. Any suggestions for a good suffix or prefix we can use to not collide with other physical tables? 

I'm thinking we should include the backup id in these snapshot tables, so ifa job doesn't clean up, we don't fail subsequent jobs, and we'd be able to easily track back to the job that didn't clean up to diagnose

busjaeger: I can think of two options:
1) reserve a fixed suffix for physical table names and add validation to the builder that the suffix is not used
2) reserve a fixed suffix (or just a separator) and add transformation logic into the builder that escapes the suffix (or separator) in physical table names if present to avoid collisions

sbabu-salesforce: going to implement option 1

PR #529, opened 2019-08-22
add profile to disable mt-dynamo test runs

sbabu-salesforce:


pom.xml
```diff
@@ -130,6 +124,12 @@
             <version>0.3.4</version>
             <scope>test</scope>
         </dependency>
+        <dependency>
+            <groupId>com.amazonaws</groupId>
+            <artifactId>DynamoDBLocal</artifactId>
+            <version>1.11.477</version>
+            <scope>test</scope>
+        </dependency>
```
busjaeger: Just an ordering change?

sbabu-salesforce: yeah, totally accidental

PR #551, opened 2019-09-08
Fix mapped key in UpdateItem, DeleteItem, and BatchGetItem

hhkwong:
There seems to be a bug in UpdateItem, DeleteItem, and BatchGetItem, when using a shared table, and the virtual table has a secondary index that includes one of the PK fields. To map the request's key, TableMapping.getItemMapper() is being used, so say if the GSI has the PK range key as its HK, the original key would have just the virtual HK and virtual RK, but the mapped key would have the physical HK, physical RK, and the physical GSI HK as well, when we should have only the physical HK and RK.


msgroi:
Great catch Helen.  These bugs are my worst nightmare.  FWIW, I did fix the bug way back(https://github.com/salesforce/mt-dynamo/commit/0d6102524e9e9d8e9cd497841d6269df2d9578a5), but (inexplicably) only in `getItem()`.  When Jeremy later implemented `batchGetItem()` he followed `getItem()` as a model so it was fixed there.

msgroi:
I think corruption was possible resulting from calls to (at least) `updateItem()` for tables that have an SI that shares fields with the table's PK.  If we're lucky, there are no such production virtual tables being used.  This should be fairly easy to determine by querying the tables definitions, compliance challenges aside.

hhkwong:
I thought the bug would only result in update/delete/batch-get requests failing when the virtual table has a secondary index that includes a table key field, so it seems like there should not be corrupted data?

PR #558, opened 2019-09-11
Make mappers into interfaces and fix querying/scanning secondary index with paging

hhkwong:
- The main purpose is to refactor the mappers, converting TableMapping, ItemMapper, ConditionMapper, QueryAndScanMapper, and RecordMapper into interfaces, in preparation for hash partitioning implementation. The current implementations are renamed RandomPartitioning*.

- While doing the refactoring, I also found that querying or scanning a secondary index with paging didn't seem to work. I think the LastEvaluatedKey wasn't being translated properly. So this PR also addresses that.

- When I made the first commit the diffs looked really hard to understand, e.g., it'd map the original ItemMapper implementation into the new ItemMapper interface, instead of the new RandomPartitioningItemMapper that is a modified version of the original imlementation. So I made a second commit ("renamed files to make diffs better") where I renamed all the interfaces so they have the prefix "I", and renamed the RandomPartitioning* classes back to their original names... So that commit would be a better diff to look at.


src/main/java/com/salesforce/dynamodbv2/mt/mappers/sharedtable/impl/ItemMapper.java
```diff
@@ -1,103 +1,35 @@
...
-                        ? fieldMapper.apply(fieldMapping, attribute)
-                        : attribute));
-            } else {
-                qualifiedItem.put(field, attribute);
-            }
-        });
-
-        return qualifiedItem;
-    }
+    Map<String, AttributeValue> applyForWrite(Map<String, AttributeValue> virtualItem);
```
hhkwong: We used to have 2 ItemMappers in TableMapping (itemMapper and keyMapper), and I'm merging them both into this ItemMapper interface. applyForWrite is for put/update requests, when we need to convert the given virtual item into the full physical item, applying the translations for both the table primary key and any secondary indexes. See applyToKeyAttributes below for the corresponding method for what used to be keyMapper.

hhkwong: I'm hoping this makes it more clear which one should be used when

src/main/java/com/salesforce/dynamodbv2/mt/mappers/sharedtable/impl/ItemMapper.java
```diff
@@ -1,103 +1,35 @@
...
-                        ? fieldMapper.reverse(fieldMapping, attribute)
-                        : attribute));
-            } else {
-                unqualifiedItem.put(field, attribute);
-            }
-        });
-        return unqualifiedItem;
-    }
+    Map<String, AttributeValue> applyToKeyAttributes(Map<String, AttributeValue> virtualItem,
+                                                     @Nullable DynamoSecondaryIndex virtualSecondaryIndex);
```
hhkwong: This is used when we are referring to an item by its table primary key only, or by a secondary index primary key if a non-null index argument is provided. Almost all uses of this are by table primary key only -- for specifying a record for update, delete, get, and batch get. The secondary index is specified only when we query or scan on an index, more specifically for translating a virtual ExclusiveStartKey into a physical one -- in this case, the start key should include all table PK attributes as well as the secondary index's PK attributes.

src/main/java/com/salesforce/dynamodbv2/mt/mappers/sharedtable/impl/MtAmazonDynamoDbBySharedTable.java
```diff
@@ -506,7 +507,13 @@ public ScanResult scan(ScanRequest scanRequest) {
...
             scanResult.setItems(items.stream().map(tableMapping.getItemMapper()::reverse).collect(toList()));
             if (scanResult.getLastEvaluatedKey() != null) {
-                scanResult.setLastEvaluatedKey(getKeyFromItem(Iterables.getLast(scanResult.getItems()), key));
+                Map<String, AttributeValue> lastItem = Iterables.getLast(scanResult.getItems());
+                Map<String, AttributeValue> lastEvaluatedKey = new HashMap<>();
+                lastEvaluatedKey.putAll(getKeyFromItem(lastItem, tableMapping.getVirtualTable().getPrimaryKey()));
+                if (scanRequest.getIndexName() != null) {
+                    lastEvaluatedKey.putAll(getKeyFromItem(lastItem, key));
+                }
+                scanResult.setLastEvaluatedKey(lastEvaluatedKey);
```
hhkwong: If we're scanning a secondary index, the last evaluated key includes not only the secondary index PK fields, but also the table PK fields

PR #565, opened 2019-09-18
fix bugs in condition mapper

hhkwong:


src/main/java/com/salesforce/dynamodbv2/mt/mappers/sharedtable/impl/ConditionMapper.java
```diff
@@ -16,14 +16,21 @@
 interface ConditionMapper {
 
     /**
-     * Maps expressions in the given request for a virtual table to a request for the physical table,
-     * based on the virtual table's primary key.
+     * Maps the update expression, condition expression, and expression attributes in the given virtual update request
+     * to the corresponding physical update request.
      */
-    void apply(RequestWrapper request);
+    void applyForUpdate(RequestWrapper request);
```
hhkwong: Handling update requests is the most difficult and has more complications: each assignment in the update expression can potentially fan out, and the update expression and filter expression may have overlapping field references. (Query requests also have possibly 2 expressions, but each can be translated independently, because the field references can't overlap -- I think.)

src/main/java/com/salesforce/dynamodbv2/mt/mappers/sharedtable/impl/ConditionMapper.java
```diff
@@ -16,14 +16,21 @@
...
 
     /**
-     * Maps expressions in the given request for a virtual table to a request for the physical table,
+     * Maps the key condition in the given virtual query request to the corresponding physical key condition,
      * based on the given virtual secondary index if it's provided, or if not, on the virtual table primary key.
      */
-    void apply(RequestWrapper request, @Nullable DynamoSecondaryIndex virtualSecondaryIndex);
+    void applyToKeyCondition(RequestWrapper request, @Nullable DynamoSecondaryIndex virtualSecondaryIndex);
```
hhkwong: Used to map key condition expression in query requests only

src/main/java/com/salesforce/dynamodbv2/mt/mappers/sharedtable/impl/ConditionMapper.java
```diff
@@ -16,14 +16,21 @@
...
+
+    /**
+     * Maps the filter expression in the given virtual request to the corresponding physical filter expression,
+     * applying the change to either the primary or secondary expression in the request wrapper.
+     */
+    void applyToFilterExpression(RequestWrapper request, boolean isPrimaryExpression);
```
hhkwong: Used to map filter/condition expressions in a put/delete/scan/query request, which in terms of RequestWrapper is primary in the case of put/delete/scan, and secondary in a query request.

src/main/java/com/salesforce/dynamodbv2/mt/mappers/sharedtable/impl/RandomPartitioningConditionMapper.java
```diff
@@ -30,143 +38,272 @@
...
+        String conditionExpression = request.getFilterExpression();
+        Set<String> virtualFieldNames = tableMapping.getAllMappingsPerField().keySet();
+
+        updateExpression = convertFieldNameLiteralsToExpressionNames(updateExpression, request, virtualFieldNames);
+        validateNotUpdatingTablePrimaryKeyFields(updateExpression, request);
+
+        if (conditionExpression != null) {
+            conditionExpression = convertFieldNameLiteralsToExpressionNames(conditionExpression, request,
+                virtualFieldNames);
+            conditionExpression = makePlaceholdersDistinct(updateExpression, conditionExpression, request);
```
hhkwong: To make our lives easier, if the condition expression has overlapping attribute name or value references with the update expression, create more placeholder references so there is no overlap

src/main/java/com/salesforce/dynamodbv2/mt/mappers/sharedtable/impl/RandomPartitioningConditionMapper.java
```diff
@@ -30,143 +38,272 @@
...
     @Override
-    public void apply(RequestWrapper request, @Nullable DynamoSecondaryIndex virtualSecondaryIndex) {
-        Collection<FieldMapping> fieldMappings;
-        if (virtualSecondaryIndex == null) {
-            fieldMappings = tableMapping.getTablePrimaryKeyFieldMappings();
+    public void applyToFilterExpression(RequestWrapper request, boolean isPrimaryExpression) {
+        String expression = isPrimaryExpression ? request.getPrimaryExpression() : request.getFilterExpression();
+
+        // for conditional / filter expressions, for each indexed field, we only need to use one of its field mappings,
+        // since we're supporting only equals expressions
```
hhkwong: This comment is outdated, before I added support for non-equals comparison operators when using non-legacy filter expressions... I'm not sure there's a bug. I'm confused because I'd expect there to be a bug if we do >, >=, <, <= comparisons on a number RK, when using binary encoding, because I thought we didn't have a binary encoding for numbers that preserves order yet...

src/main/java/com/salesforce/dynamodbv2/mt/mappers/sharedtable/impl/RandomPartitioningConditionMapper.java
```diff
@@ -30,143 +38,272 @@
...
-            Optional<String> keyFieldName = expressionAttrNames != null ? expressionAttrNames.entrySet().stream()
-                .filter(entry -> entry.getValue().equals(virtualAttrName)).map(Entry::getKey).findAny()
-                : Optional.empty(); // Optional[#field1]
-            if (keyFieldName.isPresent() && !keyFieldName.get().equals(NAME_PLACEHOLDER)) {
+    void mapFieldInConditionExpression(String expression, // "#field1 = :value"
+                                       RequestWrapper request,
+                                       FieldMapping fieldMapping) {
+        if (expression != null) {
+            String virtualField = fieldMapping.getSource().getName();
+            Set<String> fieldPlaceholders = getPlaceholdersForVirtualField(virtualField, request); // {#field1}
```
hhkwong: One of the bugs was that if there are multiple placeholders for the same field (in the same expression or across two expressions) in a request, we would only look at the first placeholder we find, and map its attribute name and possibly comparison/assignment value. So now we look at all placeholders for this virtual field.

src/main/java/com/salesforce/dynamodbv2/mt/mappers/sharedtable/impl/RandomPartitioningConditionMapper.java
```diff
@@ -30,143 +38,272 @@
...
+
+            for (String fieldPlaceholder : fieldPlaceholders) {
```
hhkwong: The diff here is weird, but given a placeholder and a field mapping, the logic is basically the same as before

src/main/java/com/salesforce/dynamodbv2/mt/mappers/sharedtable/impl/RandomPartitioningConditionMapper.java
```diff
@@ -30,143 +38,272 @@
...
+        expression = mapFieldsInConditionExpression(expression, request, fieldMappings);
+
+        if (isPrimaryExpression) {
+            request.setPrimaryExpression(expression);
         } else {
-            fieldMappings = tableMapping.getIndexPrimaryKeyFieldMappings(virtualSecondaryIndex);
+            request.setFilterExpression(expression);
         }
+    }
+
```
hhkwong: The method below is used by applyToFilterExpression() and by applyToKeyCondition() -- will add comment in code

src/main/java/com/salesforce/dynamodbv2/mt/mappers/sharedtable/impl/RandomPartitioningConditionMapper.java
```diff
@@ -30,143 +38,272 @@
...
                     String virtualValuePlaceholder = virtualValuePlaceholderOpt.get();
-                    AttributeValue virtualAttr =
+                    AttributeValue virtualValue =
                         request.getExpressionAttributeValues().get(virtualValuePlaceholder); // {S: hkValue,}
-                    AttributeValue physicalAttr =
-                        fieldMapping.isContextAware()
-                            ? fieldMapper.apply(fieldMapping, virtualAttr) // {S: ctx.virtualTable.hkValue,}
-                            : virtualAttr;
-                    request.putExpressionAttributeValue(virtualValuePlaceholder, physicalAttr);
+
```
hhkwong: Below is how we fan out from one field assignment in the virtual SET expression to possibly multiple assignments in the physical SET expression

src/main/java/com/salesforce/dynamodbv2/mt/mappers/sharedtable/impl/RandomPartitioningConditionMapper.java
```diff
@@ -30,143 +38,272 @@
...
-    private String convertFieldNameLiteralsToExpressionNamesInternal(
-        Collection<FieldMapping> fieldMappings, // source = "field", target="field"
-        String conditionExpression, // "field = :value and field2 = :value2 and field = :value3"
-        RequestWrapper request) {
-        String newConditionExpression = conditionExpression;
-        if (conditionExpression != null) {
-            AtomicInteger counter = new AtomicInteger(1);
-            for (FieldMapping fieldMapping : fieldMappings) {
-                String virtualFieldName = fieldMapping.getSource().getName(); // "field"
-                String toFind = " " + virtualFieldName + " ="; // " field ="
```
hhkwong: This wasn't working if the virtual field literal is at the beginning of the expression without a space before it

src/test/java/com/salesforce/dynamodbv2/UpdateTest.java
```diff
@@ -69,6 +79,30 @@ void update(TestArgument testArgument) {
         });
     }
 
+    @ParameterizedTest
+    @ArgumentsSource(DefaultArgumentProvider.class)
+    void updatePrimaryKeyFieldFails(TestArgument testArgument) {
```
hhkwong: New test case: can't refer to table PK fields in an update expression and throw exception consistent with single-tenant dynamo (though I got the error message from local dynamo and I'm not sure if that's necessarily the same as what AWS dynamo would give?)

src/test/java/com/salesforce/dynamodbv2/UpdateTest.java
```diff
@@ -122,56 +156,77 @@ void updateConditionalOnHkSuccess(TestArgument testArgument) {
         });
     }
 
-
     @ParameterizedTest(name = "{arguments}")
     @ArgumentsSource(DefaultArgumentProvider.class)
-    void updateConditionalOnGsiHkSuccess(TestArgument testArgument) {
+    void updateConditionalOnHkSuccessWithLiterals(TestArgument testArgument) {
```
hhkwong: New test case: same as existing test case that does conditional update, but using literals instead of placeholders

hhkwong: The diff is weird because I swapped this test case with updateConditionalOnGsiHkSuccess() so the former is adjacent to the related test case updateConditionalOnHkSuccess()...

src/test/java/com/salesforce/dynamodbv2/UpdateTest.java
```diff
@@ -122,56 +156,77 @@ void updateConditionalOnHkSuccess(TestArgument testArgument) {
...
                     testArgument.getHashKeyAttrType(),
-                    Optional.of(RANGE_KEY_S_VALUE)));
+                    Optional.empty()));
         });
     }
 
     @ParameterizedTest(name = "{arguments}")
     @ArgumentsSource(DefaultArgumentProvider.class)
-    void updateConditionalOnHkWithLiteralsSuccess(TestArgument testArgument) {
+    void updateConditionalOnGsiRkSuccess_usePlaceholder(TestArgument testArgument) {
```
hhkwong: Expanded existing test case updateConditionalOnGsiRkSuccess into 2 flavors, one using a placeholder (same as before) and one using the literal

src/test/java/com/salesforce/dynamodbv2/UpdateTest.java
```diff
@@ -208,74 +263,164 @@ void updateConditionalFail(TestArgument testArgument) {
...
-            assertEquals(TABLE3, updateItemRequest.getTableName()); // assert no side effects
+            assertEquals(tableName, updateItemRequest.getTableName()); // assert no side effects
         });
     }
 
     @ParameterizedTest
     @ArgumentsSource(DefaultArgumentProvider.class)
-    void updateHkRkTableWithGsi(TestArgument testArgument) {
-        runUpdateHkRkTableWithGsiTest(testArgument, TABLE3);
+    void updateGsiFieldAndQueryGsi(TestArgument testArgument) {
```
hhkwong: The bug that broke ZOS and what we originally talked about...

src/test/java/com/salesforce/dynamodbv2/UpdateTest.java
```diff
@@ -208,74 +263,164 @@ void updateConditionalFail(TestArgument testArgument) {
...
     }
 
     @ParameterizedTest
     @ArgumentsSource(DefaultArgumentProvider.class)
-    void updateHkRkTableWithPkFieldInGsi(TestArgument testArgument) {
-        runUpdateHkRkTableWithGsiTest(testArgument, TABLE5);
-    }
-
-    private void runUpdateHkRkTableWithGsiTest(TestArgument testArgument, String tableName) {
+    void updateFieldInMultipleGsis(TestArgument testArgument) {
```
hhkwong: New test case where we update a field that is used in multiple GSIs and requires us to fan out. At the end we query both indexes to make sure both queries work.

src/test/java/com/salesforce/dynamodbv2/testsupport/DefaultTestSetup.java
```diff
@@ -185,7 +186,7 @@ protected MtAmazonDynamoDbContextProvider getMtContext() {
         CreateTableRequestBuilder baseBuilder = CreateTableRequestBuilder.builder()
                 .withTableName(TABLE1)
                 .withAttributeDefinitions(new AttributeDefinition(HASH_KEY_FIELD, hashKeyAttrType))
-                .withKeySchema(new KeySchemaElement(HASH_KEY_FIELD, KeyType.HASH))
+                .withKeySchema(getKeySchema(HASH_KEY_FIELD, HASH))
```
hhkwong: Most changes in this file are just refactoring and don't change logic. The only logical change is that in TABLE5, I'm adding 2 new GSIs that share a common non-PK field, so I can write the test on updating a field used in multiple GSIs.

src/main/java/com/salesforce/dynamodbv2/mt/mappers/sharedtable/impl/RandomPartitioningConditionMapper.java
hhkwong: On okay I was confused because I was thinking currently mt-dynamo already supports mapping a virtual range key of type N to a physical range key of type B. But that's not true and the range key types have to be exactly the same right now. As long as we only allow RK mappings where the order is preserved, we can use any of the index FieldMappings to map comparison clauses for a given virtual field.

src/main/java/com/salesforce/dynamodbv2/mt/mappers/sharedtable/impl/ConditionMapper.java
busjaeger: By `filter expression` for updates do you mean `ConditionExpression`?

src/main/java/com/salesforce/dynamodbv2/mt/mappers/sharedtable/impl/ConditionMapper.java
busjaeger: Seems like it may be useful to revisit the `RequestWrapper` abstraction in the context of expression mapping to make this interface more explicit (in a future change).

src/main/java/com/salesforce/dynamodbv2/mt/mappers/sharedtable/impl/RandomPartitioningConditionMapper.java
busjaeger: Why is this helpful? I.e., what problem does it create to have overlapping attribute name/value placeholders?

src/main/java/com/salesforce/dynamodbv2/mt/mappers/sharedtable/impl/ConditionMapper.java
```diff
@@ -16,14 +16,21 @@
 interface ConditionMapper {
```
msgroi: I feel like some tests on these methods directly might help understand the code better.  Perhaps it's not necessary now and would be a further distraction.  I realize I should have given myself this advice when I did this in the first place.  :-|

src/main/java/com/salesforce/dynamodbv2/mt/mappers/sharedtable/impl/ConditionMapper.java
hhkwong: Yes that's what I meant

src/main/java/com/salesforce/dynamodbv2/mt/mappers/sharedtable/impl/ConditionMapper.java
hhkwong: Yeah... I was considering that too but ended up not doing it and limiting the amount of changes in this PR

src/main/java/com/salesforce/dynamodbv2/mt/mappers/sharedtable/impl/RandomPartitioningConditionMapper.java
hhkwong: Say that in the virtual update request, the update expression is `SET #fieldA = :value1`, and the condition expression is `#fieldB = :value1`, with attribute name map `{fieldA -> A, fieldB -> B}` and attribute value map `{value1 -> x}`. The virtual table has 2 GSIs defined, GSI1 has HK=A, and GSI2 has HK=C and RK=B.

Mapping the update expression, we'd apply the GSI1 FieldMapping for field A, giving us `SET #fieldA = :value1`, with name map `{fieldA -> gsi1HK}`, and value map `{value1 -> ctx/table/x}`.

Mapping the condition expression, if we did this independently without considering the update expression, we'd apply the GSI2 mapping for B, giving us `#fieldB = :value1`, with name map `{#fieldB -> gsi2RK}`, and value map `{value1 -> x}`.

But then we have conflicting value mappings for the `value1` placeholder and we'd have to deal with this somehow. I think we could have a similar scenario where the same field placeholder is used in the two expressions. Making the two expressions not share any field or value placeholders in the first place would make this easy and we can transform the two expressions independently.

busjaeger: Thanks that helps a lot!

src/main/java/com/salesforce/dynamodbv2/mt/mappers/sharedtable/impl/RandomPartitioningConditionMapper.java
```diff
@@ -175,15 +312,15 @@ static String getNextFieldPlaceholder(Map<String, String> expressionAttributeNam
...
+    static Optional<String> findVirtualValuePlaceholder(String expression, String fieldPlaceholder) {
+        if (expression == null) {
             return Optional.empty();
         }
 
-        // TODO add support for non-EQ operators
-        final Pattern p = Pattern.compile(keyFieldName + " = ([^ ,]*)");
-        final Matcher m = p.matcher(conditionExpression);
-        return m.find() ? Optional.of(m.group(1)) : Optional.empty();
+        final Pattern p = Pattern.compile(fieldPlaceholder + " *" + COMPARISON_OP_PATTERN + " *(?<val>:[a-zA-Z0-9]+)");
```
busjaeger: This is probably pretty expensive. Switching to a grammar should help here.

hhkwong:
I had ZOS's subrepo use this branch and ran all the tests (if there's a way to skip mt-dynamo tests and run just ZOS tests that'd be great), and there were 3 new tests failing, because in applyForUpdate() I was assuming there'd always be an update expression. But actually it's okay to not have an update expression, and if the key is on an existing record, it's just a no-op. Also, while looking at the ZOS tests that failed I also realized that one can call UpdateItem on a non-existent key, and that should result in a new record being created -- I think I used to know this at some point when I added support for PATCH, because it turns out the dynamo behavior is like the OData PATCH behavior, but forgot about this. So, I fixed the condition mapper so it's okay to not have any update expression, and added relevant test cases in UpdateTest so we'd catch this in mt-dynamo in the future.

sbabu-salesforce:
hey @hhkwong, just a quick comment re: ```"if there's a way to skip mt-dynamo tests and run just ZOS tests that'd be great"```, see the `make test` target in the Makefile of zos. There's a profile you can pass to `mvn test` which will skip mt-dynamo's tests, and should only run ZOS tests

PR #568, opened 2019-09-22
Hash partitioning item mapper

hhkwong:
This is the first part of supporting hash/logical partitioning, where we have a TableMapping for hash partitioning and only its ItemMapper is implemented. I added a flavor in ArgumentBuilder but it's commented out, since most tests would fail. The only tests that pass right now are the ones in GetTest, and the ones in PutTest, except for putAttributeNotExists(), which involves a condition expression.


src/main/java/com/salesforce/dynamodbv2/mt/mappers/index/DynamoSecondaryIndexMapperTrackingAssigned.java
```diff
@@ -0,0 +1,58 @@
...
+import java.util.Collection;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+import java.util.stream.Collectors;
+
+/**
+ * A secondary index mapper that keeps track of what the previous virtual-to-physical index assignments were, so that
+ * assigned physical indexes are not reused.
+ */
```
hhkwong: The purpose of this is to allow mapping multiple GSIs of a virtual table to compatible GSIs of a physical table, when the physical GSIs have the same PK types, since with hash partitioning all PKs have a binary HK and a binary RK, and the default test setup has virtual tables with multiple GSIs (each with different PK types).

src/main/java/com/salesforce/dynamodbv2/mt/mappers/sharedtable/SharedTableBuilder.java
```diff
@@ -375,9 +379,6 @@ private void setDefaults() {
         if (this.billingMode == null) {
             this.billingMode = BillingMode.PROVISIONED;
         }
-        if (secondaryIndexMapper == null) {
```
hhkwong: We were always ignoring what was set before or set here anyway, since build was always setting the index mapper after setDefaults().

src/main/java/com/salesforce/dynamodbv2/mt/mappers/sharedtable/SharedTableBuilder.java
```diff
@@ -554,18 +579,6 @@ public SharedTableBuilder withTablePrefix(String tablePrefix) {
         return this;
     }
```
hhkwong: It seems like we should not expose setters for these two, since we would always ignore what the user could have set and override them in build().

src/main/java/com/salesforce/dynamodbv2/mt/mappers/sharedtable/impl/HashPartitioningItemMapper.java
```diff
@@ -0,0 +1,359 @@
...
+import java.util.function.UnaryOperator;
+import javax.annotation.Nullable;
+
+/**
+ * Maps virtual items to physical items and vice versa for shared tables using hash partitioning. More specifically,
+ * <pre>
+ *     hashKey = tenantId + virtualTableName + hash(virtualHashKey) % numPartitions
+ *     rangeKey = virtualHashKey + virtualRangeKey
+ * </pre>
+ */
```
hhkwong: This borrows heavily from the prototype Ben did in https://github.com/salesforce/mt-dynamo/tree/salted-shared-table and the logic is very similar

src/main/java/com/salesforce/dynamodbv2/mt/mappers/sharedtable/impl/HashPartitioningItemMapper.java
```diff
@@ -0,0 +1,359 @@
...
+     * The format we use is SIGNUM + EXPONENT + D_1 + ... + D_n, where
+     * <p><ul>
+     * <li>SIGNUM: -1, 0, or 1, corresponding to negative, zero, or positive (1 byte)
+     * <li>EXPONENT: Represents exponent when number is in normalized notation (range is [-128, 127], mapped to the
+     * Dynamo range [-130, 125]; 1 byte)
+     * <li>D_1 to D_n: The n (at most 38) significand digits, from most significant to least (each digit takes 1 byte)
+     * <li>When the sign is negative, the exponent and significand digits' bits are flipped
+     * </ul></p>
+     */
+    @VisibleForTesting
```
hhkwong: Once numbers are put in normalized scientific notation, it's straightforward to sort them. E.g., it's discussed here: https://softwareengineering.stackexchange.com/questions/246021/trivially-sortable-encoding-for-arbitrary-precision-decimals.

I also looked at [how HBase does it](https://hbase.apache.org/apidocs/org/apache/hadoop/hbase/util/OrderedBytes.html), which in turn is basically [how SQLite does it](https://sqlite.org/src4/doc/trunk/www/key_encoding.wiki), but their math is a bit more complicated, and I don't know why the math works (though I'm sure Jeremy can figure it out easily).

src/main/java/com/salesforce/dynamodbv2/mt/mappers/index/DynamoSecondaryIndexMapperTrackingAssigned.java
hhkwong: We assume the order in which virtual secondary indexes are defined is always the same. If not then this is not deterministic. Ideally we'd persist the virtual-to-physical index mapping in the table description repo, and I started looking into that, but didn't do it... I thin we'll need that in order to support updating existing tables.

src/main/java/com/salesforce/dynamodbv2/mt/mappers/sharedtable/impl/MtAmazonDynamoDbBySharedTable.java
```diff
@@ -308,6 +308,8 @@ private static void validateGetItemKeysAndAttribute(KeysAndAttributes keysAndAtt
      */
     @Override
     public CreateTableResult createTable(CreateTableRequest createTableRequest) {
+        tableMappingFactory.validateCreateVirtualTableRequest(createTableRequest);
```
hhkwong: Validating that a virtual table can be created and mapped to a physical table is now done here, before we persist the virtual table description in the repo, rather than later when CRUD on the virtual table is done.

src/main/java/com/salesforce/dynamodbv2/mt/mappers/sharedtable/impl/RandomPartitioningTableMapping.java
```diff
@@ -58,19 +47,15 @@
...
-                                   DynamoSecondaryIndexMapper secondaryIndexMapper,
+    public RandomPartitioningTableMapping(DynamoTableDescription virtualTable,
+                                   DynamoTableDescription physicalTable,
+                                   UnaryOperator<DynamoSecondaryIndex> secondaryIndexMapper,
                                    MtAmazonDynamoDbContextProvider mtContext) {
-        physicalTable = lookupPhysicalTable(virtualTable, createTableRequestFactory);
-        validatePhysicalTable(physicalTable);
-        this.secondaryIndexMapper = secondaryIndexMapper;
         this.virtualTable = virtualTable;
-        validateMapping();
```
hhkwong: Validation logic has been moved out to TableMappingFactory

src/main/java/com/salesforce/dynamodbv2/mt/mappers/sharedtable/impl/TableMappingFactory.java
```diff
@@ -79,17 +87,108 @@ private void createTablesEagerly(CreateTableRequestFactory createTableRequestFac
...
-        RandomPartitioningTableMapping tableMapping = new RandomPartitioningTableMapping(virtualTableDescription,
-            createTableRequestFactory,
-            secondaryIndexMapper,
-            mtContext);
-        tableMapping.setPhysicalTable(createTableIfNotExists(tableMapping.getPhysicalTable().getCreateTableRequest()));
+    TableMapping getTableMapping(DynamoTableDescription virtualTable) {
+        DynamoTableDescription physicalTable = lookupPhysicalTable(virtualTable);
+        physicalTable = createTableIfNotExists(physicalTable.getCreateTableRequest());
+        Map<DynamoSecondaryIndex, DynamoSecondaryIndex> secondaryIndexMap =
+                virtualTableCreationValidator.validateAndGetSecondaryIndexMap(virtualTable, physicalTable);
```
hhkwong: When the index mapping is persisted, we'd get it from the repo record instead of generating it again here

src/main/java/com/salesforce/dynamodbv2/mt/mappers/sharedtable/impl/HashPartitioningItemMapper.java
```diff
@@ -0,0 +1,359 @@
...
+            if (virtualHkValue != null && virtualRkValue != null) {
+                // add physical HK
+                putPhysicalHashKey(physicalItem, virtualPK.getHashKeyType(), virtualHkValue, physicalPK);
+
+                // add physical RK
+                ByteBuffer physicalRangeKeyValue = AttributeBytesConverter.toBytes(
+                    virtualPK.getHashKeyType(), virtualHkValue, virtualPK.getRangeKeyType().get(), virtualRkValue);
+                putBytesInField(physicalItem, physicalPK.getRangeKey().get(), physicalRangeKeyValue);
+
+                // remove virtual HK and RK if needed
```
busjaeger: > if needed
It looks like it always removes them?

PR #588, opened 2019-10-06
Delete stream skip

sbabu-salesforce:


src/main/java/com/salesforce/dynamodbv2/mt/mappers/sharedtable/impl/MtAmazonDynamoDbStreamsBySharedTable.java
```diff
@@ -79,9 +81,23 @@ protected MtGetRecordsResult getAllRecords(GetRecordsRequest request, StreamArn
...
             final FieldValue<?> fieldValue = fieldValueFunction.apply(record.getDynamodb().getKeys());
             // then establish context to map physical to virtual record, i.e., remove index key prefixes
-            return mtDynamoDb.getMtContext().withContext(fieldValue.getContext(), () ->
-                mtDynamoDb.getTableMapping(fieldValue.getTableName()).getRecordMapper().apply(record)
-            );
+            return mtDynamoDb.getMtContext().withContext(fieldValue.getContext(), () -> {
+                try {
+                    return mtDynamoDb.getTableMapping(fieldValue.getTableName()).getRecordMapper().apply(record);
+
+                } catch (UncheckedExecutionException e) {
```
busjaeger: IMHO, it would be better to either throw a `ResourceNotFoundException` from `getTableMapping` (i.e., unwrap inside the method) or return `Optional`. Otherwise the implementation leaks out.

src/main/java/com/salesforce/dynamodbv2/mt/mappers/sharedtable/impl/MtAmazonDynamoDbStreamsBySharedTable.java
```diff
@@ -79,9 +81,23 @@ protected MtGetRecordsResult getAllRecords(GetRecordsRequest request, StreamArn
...
+                    // This isn't great, but we're assuming a missing virtual table entry here is a deleted virtual
+                    // table, and thus, shouldn't be mapping records to MtRecords. Instead return null, but really,
+                    // we should make sure we let deleted records flow through a stream and expire out, before removing
+                    // the virtual table entry.
+                    if (e.getCause() instanceof ResourceNotFoundException) {
+                        return null;
```
busjaeger: I wonder if returning and filtering out `null` could cause some downstream issues, because KCL now doesn't see these records, so it cannot checkpoint them. I wonder if instead we should just not convert keys or set dummy values or something.

src/main/java/com/salesforce/dynamodbv2/mt/mappers/sharedtable/impl/MtAmazonDynamoDbBySharedTable.java
```diff
@@ -414,11 +416,20 @@ public GetItemResult getItem(GetItemRequest getItemRequest) {
...
                 tableMappingFactory.getTableMapping(
-                    new DynamoTableDescriptionImpl(mtTableDescriptionRepo.getTableDescription(virtualTableName))));
+                    new DynamoTableDescriptionImpl(mtTableDescriptionRepo.getTableDescription(virtualTableName)))));
+        } catch (UncheckedExecutionException e) {
+            // This isn't great, but we're assuming a missing virtual table entry here is a deleted virtual
+            // table, and thus, shouldn't be mapping records to MtRecords. Instead return null, but really,
+            // we should make sure we let deleted records flow through a stream and expire out, before removing
+            // the virtual table entry.
+            if (e.getCause() instanceof ResourceNotFoundException) {
+                return Optional.empty();
```
busjaeger: Should we cache empty results?

sbabu-salesforce: ah good catch, will cause a lot of hitting of raw stream if we don't cache the delete markers

sbabu-salesforce: hitting of the raw description table **

busjaeger:
To simplify how we build condition, filter, and projection expressions, @hhkwong is going to change mt-dynamo to keep original index fields instead of removing them from each record.

One interesting side-effect of this change is that we can technically convert stream records without needing the table mapping by just stripping off all attribute/value pairs that are declared as index attributes in the physical table. The only thing we couldn't easily map back is `StreamRecord#keys`, since we won't know without the table mapping what the virtual hash and range key fields are. (Makes me wonder if it would be worth trying to encode that in each row somehow.)

src/main/java/com/salesforce/dynamodbv2/mt/mappers/sharedtable/impl/MtAmazonDynamoDbBySharedTable.java
```diff
@@ -414,11 +415,21 @@ public GetItemResult getItem(GetItemRequest getItemRequest) {
...
+                try {
+
+                    return Optional.of(tableMappingFactory.getTableMapping(
+                        new DynamoTableDescriptionImpl(mtTableDescriptionRepo.getTableDescription(virtualTableName))));
+                } catch (ResourceNotFoundException e) {
+                    // This isn't great, but we're assuming a missing virtual table entry here is a deleted virtual
+                    // table, and thus, shouldn't be mapping records to MtRecords. Instead return null, but really,
+                    // we should make sure we let deleted records flow through a stream and expire out, before removing
+                    // the virtual table entry.
+                    return Optional.empty();
```
horwitzj: since there seem to be a lot of `get()` calls on the return value of `getTableMapping` without `isPresent()` checks, isn't this trading `ResourceNotFoundException`s (possibly with more information) for `NoSuchElementException`s (when this catch block is reached)?

sbabu-salesforce: yes it is trading RNFE with NSEE, but in scenarios we don't expect to hit that condition, ie: from web request processing of normal CRUD. The only scenario we expect an RNFE is following a delete of a virtual table, and a subsequent stream trying to process data from that deleted virtual table

PR #595, opened 2019-10-08
silence noisy logging

sbabu-salesforce:


sbabu-salesforce:
Every time a virtual table description is loaded (cache miss), we log details about the table. This is especially noisy during zos dyno restarts.

src/main/java/com/salesforce/dynamodbv2/mt/mappers/sharedtable/impl/TableMappingFactory.java
```diff
@@ -189,14 +189,14 @@ TableMapping getTableMapping(DynamoTableDescription virtualTable) {
...
-        LOG.info("created virtual to physical table mapping: " + tableMapping.toString());
+        LOG.debug("created virtual to physical table mapping: " + tableMapping.toString());
         return tableMapping;
     }
 
     private DynamoTableDescriptionImpl createTableIfNotExists(CreateTableRequest physicalTable) {
         // does not exist, create
         if (getTableDescription(physicalTable.getTableName()).isPresent()) {
-            LOG.info(format("using existing physical table %s", physicalTable.getTableName()));
+            LOG.debug(format("using existing physical table %s", physicalTable.getTableName()));
```
busjaeger: I don't think the problem is that we're logging at info level. The problem is that we are calling this method all the time. I am trying to fix that with #613 .

src/main/java/com/salesforce/dynamodbv2/mt/mappers/sharedtable/impl/TableMappingFactory.java
```diff
@@ -189,14 +189,14 @@ TableMapping getTableMapping(DynamoTableDescription virtualTable) {
                 virtualTableCreationValidator.validateAndGetSecondaryIndexMap(virtualTable, physicalTable);
         TableMapping tableMapping = partitioningStrategy.createTableMapping(virtualTable, physicalTable,
             secondaryIndexMap::get, mtContext);
-        LOG.info("created virtual to physical table mapping: " + tableMapping.toString());
+        LOG.debug("created virtual to physical table mapping: " + tableMapping.toString());
```
busjaeger: We should add a guard `if (LOG.isDebugEnabled)` and/or use log templates `LOG.debug("created virtual to physical table mapping: {}", tableMapping.toString());`, so that we don't construct `String`s unnecessarily when debug logging is not enabled.

src/main/java/com/salesforce/dynamodbv2/mt/mappers/sharedtable/impl/TableMappingFactory.java
sbabu-salesforce: Regardless of how we cache this create table lookup, every restart of zos involves wiping the cache, and reloading across all tenants. The issue I noticed was a barrage of these on zos startup, which I dont think that PR addresses.

busjaeger: I think this message should only be logged once on restart with the changes in PR #613.

PR #599, opened 2019-10-10
support byte buffer backups and restore to different tenant

sbabu-salesforce:
This change was prompted by writing an integration test from ZOS that: 
- writes a dummy entity set to a dummy scope
- takes a mt backup of the shared table
- restore the scopes' data to a different scope

It's a test case that gives me confidence to start actually taking periodic backups, as we implement the different types of restore support we're looking for.


src/main/kotlin/com/salesforce/dynamodbv2/mt/backups/GsonByteBufferTypeAdapter.kt
```diff
@@ -0,0 +1,24 @@
+/* Copyright (c) 2019, Salesforce.com, Inc.
```
sbabu-salesforce: [shamelessly] plagiarized from https://github.com/oktolab/gson-utils/blob/master/src/main/java/br/com/oktolab/gson/adapter/GsonByteBufferTypeAdapter.java

src/main/java/com/salesforce/dynamodbv2/mt/mappers/sharedtable/impl/MtAmazonDynamoDbBySharedTable.java
```diff
@@ -697,10 +710,14 @@ public DeleteBackupResult deleteBackup(DeleteBackupRequest deleteBackupRequest)
         if (backupManager.isPresent()) {
```
sbabu-salesforce: validate if mt context is not specified here (no tenant specific delete backup supported yet)

src/main/kotlin/com/salesforce/dynamodbv2/mt/backups/GsonByteBufferTypeAdapter.kt
```diff
@@ -0,0 +1,24 @@
...
+ * SPDX-License-Identifier: BSD-3-Clause.
+ * For full license text, see LICENSE file in the repo root or https://opensource.org/licenses/BSD-3-Clause.
+ */
+package com.salesforce.dynamodbv2.mt.backups
+
+import com.google.gson.*
+import java.lang.reflect.Type
+import java.nio.ByteBuffer
+import java.util.*
+
```
sbabu-salesforce: add comment

src/main/java/com/salesforce/dynamodbv2/mt/mappers/sharedtable/impl/MtAmazonDynamoDbBySharedTable.java
```diff
@@ -645,14 +647,14 @@ public ListBackupsResult listBackups(ListBackupsRequest listBackupsRequest) {
     public RestoreTableFromBackupResult restoreTableFromBackup(
         RestoreTableFromBackupRequest restoreTableFromBackupRequest) {
         if (backupManager.isPresent()) {
-            //validate we have a backup for this tenant-table and we're under the same context
+            if (!getMtContext().getContextOpt().isPresent()) {
```
busjaeger: Minor: `!isPresent() == isEmpty()`

src/main/java/com/salesforce/dynamodbv2/mt/mappers/sharedtable/impl/MtAmazonDynamoDbBySharedTable.java
```diff
@@ -645,14 +647,14 @@ public ListBackupsResult listBackups(ListBackupsRequest listBackupsRequest) {
...
+            }
+            //validate we have a backup for this tenant-table
             try {
-                TenantTableBackupMetadata tenantBackupMetadata = backupManager.get()
-                    .getTenantTableBackupFromArn(restoreTableFromBackupRequest.getBackupArn());
-                checkArgument(tenantBackupMetadata.getTenantId().equals(getMtContext().getContext()),
-                    "Current context does not match ARN context");
-            } catch (IllegalArgumentException e) {
-                throw new MtBackupException("Error restoring table, invalid input.", e);
+                describeBackup(new DescribeBackupRequest().withBackupArn(restoreTableFromBackupRequest.getBackupArn()));
```
busjaeger: Why is it validating this explicitly up-front (as opposed to throwing an exception during restore when discovering a missing s3 file or something)?

src/main/kotlin/com/salesforce/dynamodbv2/mt/backups/MtBackupAwsAdaptor.kt
```diff
@@ -51,11 +51,39 @@ class MtBackupAwsAdaptor {
...
+            throw IllegalArgumentException("$backupArn does not include tenant-table specifier.")
+        }
+        val tenantTableParts = backupArn.split(':')
+        return TenantTableBackupMetadata(tenantTableParts[0], tenantTableParts[1], tenantTableParts[2])
+    }
+
+    private fun isTenantTableArn(backupArn: String): Boolean = backupArn.split(':').size == 3
+
+    fun getBackupArnForTenantTableBackup(tenantTable: TenantTableBackupMetadata): String =
+            "${tenantTable.backupName}:${tenantTable.tenantId}:${tenantTable.virtualTableName}"
```
busjaeger: Have you considered using the same format we use for virtual stream arns, i.e., appending virtual table name and tenant id to the end of the arn?

For backup something like:
`arn:${Partition}:dynamodb:${Region}:${Account}:table/${TableName}/backup/${BackupName}/context/${tenantId}/tenantTable/${tenantTable}`

(source: https://docs.aws.amazon.com/IAM/latest/UserGuide/list_amazondynamodb.html#amazondynamodb-resources-for-iam-policies)

sbabu-salesforce: will take a stab trying to align with that... I think it might make sense to use the tenant table backup metadata json file ARN for the s3 resource, so its actually a useful ARN vs arbitrarily aligned

src/main/java/com/salesforce/dynamodbv2/mt/mappers/sharedtable/impl/MtAmazonDynamoDbBySharedTable.java
sbabu-salesforce: I wanna say it was an NPE when the backup ARN wasn't found, but that's not showing up when the ARN comes out of listBackups (vs ZOS duplicating ARN construction logic). Just removed this snippet

src/main/kotlin/com/salesforce/dynamodbv2/mt/backups/MtBackupAwsAdaptor.kt
```diff
@@ -51,11 +51,39 @@ class MtBackupAwsAdaptor {
...
+        return TenantTableBackupMetadata(tenantTableParts[0], tenantTableParts[1], tenantTableParts[2], tenantTableParts[3])
+    }
+
+    private fun isTenantTableArn(backupArn: String): Boolean = backupArn.split(':').size == 4
+
+    fun getBackupArnForTenantTableBackup(tenantTable: TenantTableBackupMetadata): String =
+            "${tenantTable.s3BucketName}:${tenantTable.backupName}:${tenantTable.tenantId}:${tenantTable.virtualTableName}"
+
+    fun getBackupArnForTenantTableBackup(tenantTable: TenantBackupMetadata): String =
+            "${tenantTable.s3BucketName}:${tenantTable.backupName}:${tenantTable.tenantTable.tenantName}:${tenantTable.tenantTable.virtualTableName}"
```
busjaeger: It seems a bit weird that we're returning an s3 arn instead of a dynamodb backup arn, but I guess either way we're not sticking with their format, so it's ok?

sbabu-salesforce: its not even an s3 arn, its totally its own thing.. I figured the s3 bucket is important information if/when we try restoring backups from one app into another. In the shorter term, its also import to distinguish data backups from metadata backups with the zos change creating and restoring backups from different buckets.

I imagine wanting to add logic at some point in the future of making sure to download the backup from the bucket in the arn, vs what the mt-dynamo backup manager is currently configured to.

PR #604, opened 2019-10-15
Support queries and scans when using hash partitioning

hhkwong:


src/main/java/com/salesforce/dynamodbv2/mt/mappers/sharedtable/impl/AbstractQueryAndScanMapper.java
```diff
@@ -0,0 +1,292 @@
...
+    AbstractQueryAndScanMapper(DynamoTableDescription virtualTable,
+                               Function<String, RequestIndex> indexLookup,
+                               ConditionMapper conditionMapper,
+                               ItemMapper itemMapper) {
+        this.virtualTable = virtualTable;
+        this.indexLookup = indexLookup;
+        this.conditionMapper = conditionMapper;
+        this.itemMapper = itemMapper;
+    }
+
```
hhkwong: A lot of the logic here was pulled out from RandomPartitioningQueryAndScanMapper, e.g., converting legacy request parameters into key condition / filter expressions.

src/main/java/com/salesforce/dynamodbv2/mt/mappers/sharedtable/impl/AbstractQueryAndScanMapper.java
```diff
@@ -0,0 +1,292 @@
...
+        // map filter expression if there is one
+        QueryRequestWrapper filterExpressionRequest = new QueryRequestWrapper(queryRequest,
+            queryRequest::getFilterExpression, queryRequest::setFilterExpression);
+        applyToFilterExpression(filterExpressionRequest);
+
+        // map exclusive start key if there is one
+        applyExclusiveStartKey(filterExpressionRequest, requestIndex);
+    }
+
```
hhkwong: The interface now has executeScan() instead of an apply() that takes a ScanRequest, since for hash partitioning does things differently and runs physical QueryRequests instead.

src/main/java/com/salesforce/dynamodbv2/mt/mappers/sharedtable/impl/AbstractQueryAndScanMapper.java
```diff
@@ -0,0 +1,292 @@
...
+        Map<String, AttributeValue> getExclusiveStartKey();
+
+        void setExclusiveStartKey(Map<String, AttributeValue> exclusiveStartKey);
+
+        String getIndexName();
+
+        void setIndexName(String indexName);
+    }
+
```
hhkwong: I also did some refactoring of RequestWrapper, so that each wrapper is used to get/set at most 1 expression, instead of possibly 2 expressions (primaryExpression and filterExpression). This seems less confusing to me.

src/main/java/com/salesforce/dynamodbv2/mt/mappers/sharedtable/impl/HashPartitioningConditionMapper.java
```diff
@@ -0,0 +1,367 @@
...
+     *     <li>"#vhk = :vhk", no vrk ==> "#prk = :vhk“
+     *     <li>"#vhk = :vhk and #vrk = :vrk" ==> "#prk = :vhk+vrk“
+     *     <li>"#vhk = :vhk", has vrk ==> "#prk BETWEEN :vhk AND :vhk+FFFFFF)"
+     *     <li>"#vhk = :vhk and #vrk < :vrk" ==> "#prk BETWEEN :vhk AND :vhk+(vrk-1)+FFFFFF“
+     *     <li>"#vhk = :vhk and #vrk <= :vrk" ==> "#prk BETWEEN :vhk AND :vhk+vrk“
+     *     <li>"#vhk = :vhk and #vrk > :vrk" ==> "#prk BETWEEN :vhk+vrk+0 AND :vhk+FFFFFF“
+     *     <li>"#vhk = :vhk and #vrk >= :vrk" ==> "#prk BETWEEN :vhk+vrk AND :vhk+FFFFFF“
+     *     <li>"#vhk = :vhk and #vrk BETWEEN :vrk1 AND :vrk2" ==> "#prk BETWEEN :vhk+vrk1 AND :vhk+vrk2“
+     * </ul>
+     */
```
hhkwong: The main query key condition expression mapping logic is here

src/main/java/com/salesforce/dynamodbv2/mt/mappers/sharedtable/impl/HashPartitioningItemMapper.java
```diff
@@ -110,60 +96,24 @@ private void addPhysicalKeys(Map<String, AttributeValue> virtualItem, Map<String
...
                 // add physical RK
-                ByteBuffer physicalRangeKeyValue = AttributeBytesConverter.toBytes(
-                    virtualPK.getHashKeyType(), virtualHkValue, virtualPK.getRangeKeyType().get(), virtualRkValue);
-                putBytesInField(physicalItem, physicalPK.getRangeKey().get(), physicalRangeKeyValue);
-
-                // remove virtual HK and RK if needed
-                physicalItem.remove(virtualPK.getHashKey());
-                physicalItem.remove(virtualPK.getRangeKey().get());
+                physicalItem.put(physicalPK.getRangeKey().get(),
+                    keyMapper.toPhysicalRangeKey(virtualPK, virtualHkValue, virtualRkValue));
```
hhkwong: We now maintain all the original virtual PK/index field values, which makes our lives a lot easier, since we don't have to translate filter & projection expressions.

src/main/java/com/salesforce/dynamodbv2/mt/mappers/sharedtable/impl/HashPartitioningItemMapper.java
```diff
@@ -172,188 +122,40 @@ private int getPrimitiveValueHashCode(ScalarAttributeType type, AttributeValue v
...
+                    virtualPK.getRangeKeyType().get(), physicalRkBytes);
+                item.put(virtualPK.getHashKey(), virtualValues[0]);
+                item.put(virtualPK.getRangeKey().get(), virtualValues[1]);
             } else {
-                item.put(virtualPK.getHashKey(), AttributeBytesConverter.fromBytes(virtualPK.getHashKeyType(),
-                    physicalRkValue.getB()));
-            }
-        }
-    }
-
```
hhkwong: The logic for converting virtual HK and RK values into physical HK and RK bytes is now in HashPartitioningKeyMapper (and its inner classes)

src/main/java/com/salesforce/dynamodbv2/mt/mappers/sharedtable/TablePartitioningStrategy.java
```diff
@@ -52,88 +44,14 @@
      */
     void validateCompatiblePrimaryKeys(PrimaryKey virtualPrimaryKey, PrimaryKey physicalPrimaryKey);
 
+    /**
+     * Parses a given physical hash key value into the record's MT context and virtual table name.
+     */
+    MtContextAndTable toContextAndTable(ScalarAttributeType physicalHashKeyType, AttributeValue physicalHashKeyValue);
+
```
hhkwong: To scan all tenants, we need a way to parse the MT context and virtual table name from a given physical record

src/main/java/com/salesforce/dynamodbv2/mt/mappers/sharedtable/impl/HashPartitioningKeyMapper.java
```diff
@@ -0,0 +1,307 @@
...
+        private static final int MAX_EXPONENT = 125;
+
+        static byte[] encode(BigDecimal bigDecimal) {
+            bigDecimal = bigDecimal.stripTrailingZeros();
+            int precision = bigDecimal.precision();
+            byte[] byteArray = new byte[2 + precision];
+
+            // first byte represents the signum
+            byteArray[0] = toSignumByte(bigDecimal.signum());
+
```
hhkwong: Moved from HashPartitioningItemMapper, with an important fix to treat bytes as unsigned, since that's how Dynamo stores them.

src/main/java/com/salesforce/dynamodbv2/mt/mappers/sharedtable/impl/MtAmazonDynamoDbBySharedTable.java
```diff
@@ -813,16 +792,22 @@ private DeleteTableResult deleteTableInternal(DeleteTableRequest deleteTableRequ
...
+                    break;
+                }
+                scanRequest.withExclusiveStartKey(scanResult.getLastEvaluatedKey());
             }
-            log.warn("truncation of " + scanResult.getItems().size() + " items from table=" + tableName
-                + (scanResult.getLastEvaluatedKey() == null
-                ? " complete. "
-                : "but data may have been dropped to the floor, beware."));
-
+            log.warn("truncation of " + deletedCount + " items from table=" + tableName + " complete.");
```
hhkwong: Truncating all records of a a deleted table instead of only the first page of a scan

src/main/java/com/salesforce/dynamodbv2/mt/mappers/sharedtable/impl/RandomPartitioningConditionMapper.java
```diff
@@ -297,21 +289,6 @@ static String makePlaceholdersDistinct(String primaryExpression, String expressi
...
-    static String getNextPlaceholder(Map<String, ?> existingPlaceholders, String prefix) {
-        int counter = 1;
-        String placeholderCandidate = prefix + counter;
-        while (existingPlaceholders != null && existingPlaceholders.containsKey(placeholderCandidate)) {
-            counter++;
-            placeholderCandidate = prefix + counter;
-        }
-        return placeholderCandidate;
-    }
-
```
hhkwong: Moved out as common util method

src/main/java/com/salesforce/dynamodbv2/mt/mappers/sharedtable/impl/RandomPartitioningQueryAndScanMapper.java
```diff
@@ -7,121 +7,91 @@
...
+        this.tablePrimaryKeyFieldMappings = tablePrimaryKeyFieldMappings;
+        this.indexPrimaryKeyFieldMappings = indexPrimaryKeyFieldMappings;
     }
 
     @Override
-    public void apply(QueryRequest queryRequest) {
-        validateQueryRequest(queryRequest);
-        apply(new QueryRequestWrapper(queryRequest), true);
-    }
+    ScanResult executeScanInternal(AmazonDynamoDB amazonDynamoDB, ScanRequest scanRequest) {
```
hhkwong: Logic for random partitioning scan is basically still the same. Changes are due to refactoring.

src/test/java/com/salesforce/dynamodbv2/BatchGetTest.java
```diff
@@ -39,7 +38,8 @@
 class BatchGetTest {
 
     @ParameterizedTest(name = "{arguments}")
-    @ArgumentsSource(DefaultArgumentProviderForTable1.class)
+    @ArgumentsSource(DefaultArgumentProvider.class)
+    @DefaultArgumentProviderConfig(tables = {TABLE1})
```
hhkwong: Introduced a new annotation for tests using DefaultArgumentProvider, so we can specify which test tables to set up in a nicer way, and also so that we can specify Dynalite to be used instead of DynamoDB Local in select test cases.

src/test/java/com/salesforce/dynamodbv2/QueryTest.java
```diff
@@ -120,149 +138,186 @@ void queryWithKeyConditionsEq(TestArgument testArgument) {
...
-    @ParameterizedTest(name = "{arguments}")
-    @ArgumentsSource(DefaultArgumentProviderForTable4.class)
-    void queryWithKeyConditionsGeAlmostAsLow(TestArgument testArgument) {
-        queryWithKeyConditionsComparisonOperatorInner(GE,
-            testArgument,
-            String.valueOf(RANGE_KEY_N_MIN + 1),
-            IntStream.rangeClosed(RANGE_KEY_N_MIN + 1, RANGE_KEY_N_MAX)
-                .mapToObj(String::valueOf)
-                .collect(Collectors.toSet()));
+    private void runHkRkTableKeyConditionTest(TestArgument testArgument, boolean useLegacy) {
```
hhkwong: We used to have better coverage for queries using legacy key conditions but less coverage for those using key condition expressions, so this fixes that. Also thought about keeping these as different test cases, but with Dynalite already making these tests run slower, and since the meat of test is read-only, here I'm the test cases combined, so we don't have to set up test data once for every case.

src/test/java/com/salesforce/dynamodbv2/QueryTest.java
```diff
@@ -312,71 +369,8 @@ void queryUsingAttributeNameLiterals(TestArgument testArgument) {
     }
 
     @ParameterizedTest(name = "{arguments}")
```
hhkwong: Now covered by queryHkRkWithKeyConditions

src/main/java/com/salesforce/dynamodbv2/mt/mappers/sharedtable/impl/HashPartitioningKeyMapper.java
```diff
@@ -0,0 +1,297 @@
...
+    @VisibleForTesting
+    static final String DELIMITER = ".";
+
+    private static final String DELIMITER_PATTERN = "\\.";
+
+    static class HashPartitioningKeyPrefixFunction {
+
+        static AttributeValue toPhysicalHashKey(String context, String virtualTableName, int bucket) {
+            String stringValue = String.join(DELIMITER, context, virtualTableName, String.valueOf(bucket));
```
busjaeger: I wonder if it would make sense to serialize these values using an actual binary encoding, i.e., encode bucket as four byte number (or maybe 2 bytes, depending on how big we think the max will get) and context and virtual table name as variable-length strings (delimited & escaped). That would make it more space efficient, especially, if we allow different context types in the future (e.g., int).

If we stick with a string, we could probably store it as such to make it easier to work with the system (since dynamodb internally just converts the string into a byte array the same way).

hhkwong: Yeah that makes sense. So for example, some format like bucket + contextBytesLength + contextBytes + virtualTableNameBytes?
It can also make it easier to figure out only the bucket number given a physical hash key, though I'm not sure if we need that.

src/main/java/com/salesforce/dynamodbv2/mt/mappers/sharedtable/impl/HashPartitioningConditionMapper.java
```diff
@@ -0,0 +1,367 @@
...
+                    throw new IllegalArgumentException("Key condition expression contains non-key field: " + fieldName);
+                }
+            }
+            return visitChildren(keyConditionPart);
+        }
+
+        private AttributeValue removeValuePlaceholder(String placeholder) {
+            AttributeValue value = null;
+            if (valuePlaceholders != null) {
+                value = valuePlaceholders.remove(placeholder);
```
busjaeger: What happens if the map passed in by the client of mt-dynamo contains placeholders that do not appear in expressions?

src/main/java/com/salesforce/dynamodbv2/mt/mappers/sharedtable/impl/HashPartitioningConditionMapper.java
```diff
@@ -0,0 +1,367 @@
...
+            String literal = null;
+            if (fieldPlaceholders != null) {
+                literal = fieldPlaceholders.remove(placeholder);
+            }
+            checkArgument(literal != null, "Referenced attribute name does not exist: %s", placeholder);
+            return literal;
+        }
+    }
+
+    private static ComparisonOperator getComparatorType(String comparator) {
```
busjaeger: Antlr doesn't do this for you?

src/main/java/com/salesforce/dynamodbv2/mt/mappers/sharedtable/impl/HashPartitioningConditionMapper.java
```diff
@@ -0,0 +1,367 @@
...
+                    checkArgument(hkValue == null, "Multiple hash key conditions");
+                    ComparatorContext comparator = keyConditionPart.comparator();
+                    checkArgument(comparator != null && comparator.getText().equals("="),
+                        "Hash key condition must have operator '='");
+                    String valuePlaceholder = keyConditionPart.literal(0).getText();
+                    hkValue = removeValuePlaceholder(valuePlaceholder);
+                } else if (primaryKey.getRangeKey().isPresent() && primaryKey.getRangeKey().get().equals(fieldName)) {
+                    checkArgument(rkCondition.isEmpty(), "Multiple range key conditions");
+                    ComparatorContext comparator = keyConditionPart.comparator();
+                    ComparisonOperator op = comparator != null ? getComparatorType(comparator.getText()) : BETWEEN;
```
busjaeger: Why does it default to `BETWEEN`?

busjaeger: Oh was thinking of random partitioning ... makes sense that it defaults to between.

hhkwong: This is because in the ANTLR grammar definition, we have a rule called `comparator` that defines the single-value comparators (i.e., `=`, `<>`, `>`, `>=`, `<`, `<=`), whereas `BETWEEN` is defined as a separate lexer rule, since its syntax is different (`id comparator literal` vs. `id BETWEEN literal AND literal`). ComparatorContext here corresponds to the `comparator` rule. And based on the way the `keyConditionExpression` rule is defined, when there is an `id`, and there is no `comparator`, there should be a `BETWEEN` token. But maybe it'd be better if I check that explicitly, instead of defaulting to BETWEEN like this.

src/main/java/com/salesforce/dynamodbv2/mt/mappers/sharedtable/impl/HashPartitioningConditionMapper.java
```diff
@@ -0,0 +1,367 @@
...
+        }
+
+        private BetweenRange getBetweenRange(ComparisonOperator comparisonOperator, List<AttributeValue> vrkValues) {
+            ScalarAttributeType virtualRkType = virtualPk.getRangeKeyType().get();
+            switch (comparisonOperator) {
+                case LT:
+                    // "#vhk = :vhk and #vrk < :vrk" ==> "#prk BETWEEN :vhk AND :vhk+(vrk-1)+FFFFFF“
+                    ByteBuffer rkBytes = PhysicalRangeKeyBytesConverter.toBytes(virtualRkType, vrkValues.get(0));
+                    byte[] rkByteArray = Arrays.copyOf(rkBytes.array(), rkBytes.limit());
+                    subtractOneFromByteArray(rkByteArray);
```
busjaeger: This is computing the largest value that is smaller than the specified virtual hash key, since BETWEEN is inclusive, correct?

src/main/java/com/salesforce/dynamodbv2/mt/mappers/sharedtable/impl/HashPartitioningConditionMapper.java
hhkwong: Oh good point... yeah I think this is a bug. This value can also appear in the query's filter expression, in which case we can't remove it...

Or are you asking what the underlying Dynamo would do if there is a field or value placeholder that doesn't appear in any expressions in the request? At least according to behavior I've seen from DynamoDB Local, the request will be rejected in that case.

src/main/java/com/salesforce/dynamodbv2/mt/mappers/sharedtable/impl/HashPartitioningConditionMapper.java
hhkwong: The return value here is a com.amazonaws.services.dynamodbv2.model.ComparisonOperator

src/main/java/com/salesforce/dynamodbv2/mt/mappers/sharedtable/impl/HashPartitioningConditionMapper.java
hhkwong: Right that's what I'm trying to do

src/main/java/com/salesforce/dynamodbv2/mt/mappers/sharedtable/impl/HashPartitioningConditionMapper.java
busjaeger: Great javadoc, thanks!

src/main/java/com/salesforce/dynamodbv2/mt/mappers/sharedtable/impl/HashPartitioningConditionMapper.java
```diff
@@ -0,0 +1,367 @@
...
+                case LE:
+                    // "#vhk = :vhk and #vrk <= :vrk" ==> "#prk BETWEEN :vhk AND :vhk+vrk“
+                    return new BetweenRange(
+                        keyMapper.toPhysicalRangeKey(virtualPk, virtualHkValue, EMPTY_BYTE_ARRAY),
+                        keyMapper.toPhysicalRangeKey(virtualPk, virtualHkValue, vrkValues.get(0)));
+                case GT:
+                    // #vhk = :vhk and #vrk > :vrk" ==> "#prk BETWEEN :vhk+vrk+0 AND :vhk+FFFFFF“
+                    rkBytes = PhysicalRangeKeyBytesConverter.toBytes(virtualRkType, vrkValues.get(0));
+                    byte[] lowerBoundBytes = Arrays.copyOf(rkBytes.array(), rkBytes.limit() + 1);
+                    lowerBoundBytes[lowerBoundBytes.length - 1] = 0;
```
busjaeger: What if the vhk+vrk already maxes out the prk length?

src/main/java/com/salesforce/dynamodbv2/mt/mappers/sharedtable/impl/HashPartitioningConditionMapper.java
busjaeger: So for this to be correct, there must not exist a value `v`, such that `vhk+(vrk-1)+FFFFF < v < vhk+vrk`, right? Can you explain why that's the case? (Not doubting that this is correct, just making sure I understand it.)

src/main/java/com/salesforce/dynamodbv2/mt/mappers/sharedtable/impl/HashPartitioningConditionMapper.java
hhkwong: Good point -- should check for that and handle it differently. If vhk+vrk is already at vhk+FFFFFF, then throw; otherwise, lowerBound is vhk+(vrk+1).

src/main/java/com/salesforce/dynamodbv2/mt/mappers/sharedtable/impl/HashPartitioningKeyMapper.java
busjaeger: We could use something like the ZOS encoding:
https://git.soma.salesforce.com/salesforce-zero/zero-object-service/blob/a5d7a2203b074e09fc6659c3eeb567bbef30eec3/zos/src/main/java/com/salesforce/zos/dynamo/processors/Codec.java

I wasn't necessarily proposing to change the order in which we store the three fields, though I suppose we could do that (not sure it makes a difference).

So far, we've been restricting the type (String) and characters for context and virtual table name, so that we can keep a consistent interface with `byTable`. Which is why I think using `.` as the separator works. However, if we move away from supporting both implementations and introduce explicit multi-tenant tables, we may would want to reconsider that, i.e., allow using different context types, such as integer, or other characters. In that case, we'd need some kind of escape/separator or length approach.

src/main/java/com/salesforce/dynamodbv2/mt/mappers/sharedtable/impl/HashPartitioningQueryAndScanMapper.java
```diff
@@ -0,0 +1,108 @@
...
+        // the initial bucket.
+        String physicalHkFieldPlaceholder = MappingUtils.getNextFieldPlaceholder(queryRequestWrapper);
+        String physicalHkValuePlaceholder = MappingUtils.getNextValuePlaceholder(queryRequestWrapper);
+        queryRequest.withKeyConditionExpression(physicalHkFieldPlaceholder + " = " + physicalHkValuePlaceholder)
+            .addExpressionAttributeNamesEntry(physicalHkFieldPlaceholder, physicalPk.getHashKey())
+            .addExpressionAttributeValuesEntry(physicalHkValuePlaceholder, keyMapper.toPhysicalHashKey(bucket));
+
+        // keep moving forward pages until we find at least one record or reach end
+        QueryResult queryResult;
+        while ((queryResult = amazonDynamoDB.query(queryRequest)).getItems().isEmpty()) {
```
busjaeger: So we query one bucket and pass the user-specified limit down?

The doc says: `A single Scan operation reads up to the maximum number of items set (if using the Limit parameter) or a maximum of 1 MB of data and then apply any filtering to the results using FilterExpression.`

It seems like ideally, we'd implement that contract (for random partitioning as well), but we probably don't have enough data returned from the `ScanResult` to make that distinction, right? (We are running into the same problem with streams.)

src/main/java/com/salesforce/dynamodbv2/mt/mappers/sharedtable/impl/HashPartitioningConditionMapper.java
hhkwong: Thanks for catching this! This is actually incorrect and I'm fixing it now. The algorithm should be: if the given VRK's least significant byte is 0, then strip the 0; otherwise, subtract 1 from the least significant byte.

For the case where the least significant byte is non-zero, it suffices to show that there doesn't exist a value `v` such that `(vrk-1)+FFFFFF < v < vrk` (where `vrk-1` is what we get when we subtract 1 from the last element). Suppose there is such a `v`. It's clear that all 3 sequences must share the same subsequence up to index `n-1` (1-indexed), where `n` is the length of `vrk`. At index `n`, since `v < vrk` and `vrk` has no more elements, it must be that `v[n] < vrk[n]`. So `v[n] <= vrk[n] - 1`, which means `v[n] = (vrk-1)[n]`. Then it's clear that if we fill the bytes after `(vrk-1)[n]` with the max unsigned byte, there isn't a way for `(vrk-1)+FFFFFF < v` to be true.

src/main/java/com/salesforce/dynamodbv2/mt/mappers/sharedtable/impl/HashPartitioningConditionMapper.java
hhkwong: Fixing this now. I now have this such that if the length of vhk+vrk is at the maximum capacity, then find the least signficant i such that vrk[i] is less than the max unsigned byte. Add 1 to vrk[i], and return vrk[0, i].

src/main/java/com/salesforce/dynamodbv2/mt/mappers/sharedtable/impl/HashPartitioningQueryAndScanMapper.java
hhkwong: Yeah... I don't think there's anything in ScanResult / QueryResult telling us that. We could estimate the size of each returned item and aggregate the sizes, but I don't know how inaccurate that would be.

src/main/java/com/salesforce/dynamodbv2/mt/mappers/sharedtable/impl/HashPartitioningKeyMapper.java
hhkwong: Cool I see. Since we're already storing composite keys in the physical RK using the length approach, and the lengths of context and table can be stored in shorts (can't be more than 2048), maybe we can also use the length approach here? (I find it simpler than escaping.)

src/main/java/com/salesforce/dynamodbv2/mt/mappers/sharedtable/impl/HashPartitioningConditionMapper.java
```diff
@@ -0,0 +1,426 @@
...
+            if (valuePlaceholders != null) {
+                // keep the value placeholder if it's also in the filter expression; otherwise remove it
+                value = isPlaceholderInFilterExpression(placeholder)
+                    ? valuePlaceholders.get(placeholder)
+                    : valuePlaceholders.remove(placeholder);
+            }
+            checkArgument(value != null, "Referenced attribute value does not exist: %s", placeholder);
+            return value;
+        }
+
```
hhkwong: There was a bug both here and also in random partitioning, where a query request has a key condition expression and filter expression, and there's a value placeholder that appears in both expressions. Before I was always removing the value placeholder here and that's wrong if it also appears in the filter expression. For checking whether whether a placeholder appears in the filter expression is based on patterns/matchers because it seems overkill to define the filter expression grammar just for this purpose? Though we actually did start defining one already, though it's incomplete.

hhkwong: Actually now I'm thinking I'll define the full grammar in ANTLR and use its parser instead. It seems bad to compile patterns so often, though I don't have a good idea of what the cost would be.

hhkwong: OK then I realized I don't need to define the full parser rule for filter/condition expressions and can just use lexer rules. So that's how we get the value placeholders from the filter expression now.

src/main/java/com/salesforce/dynamodbv2/mt/mappers/sharedtable/impl/HashPartitioningKeyMapper.java
busjaeger: Sounds good

src/main/java/com/salesforce/dynamodbv2/mt/mappers/sharedtable/impl/HashPartitioningConditionMapper.java
busjaeger: Thanks, that makes sense.

PR #609, opened 2019-10-17
disallow populating scan tenant reserved column keys

sbabu-salesforce:


src/main/kotlin/com/salesforce/dynamodbv2/mt/mappers/sharedtable/impl/MtSharedTableBackupManager.kt
```diff
@@ -453,6 +453,8 @@ open class MtSharedTableBackupManager(
                 val row = scanResult.items[i]
                 val tenant = row.get(sharedTableMtDynamo.scanTenantKey)!!.s
                 val virtualTable = row.get(sharedTableMtDynamo.scanVirtualTableKey)!!.s
+                row.remove(sharedTableMtDynamo.scanTenantKey)
```
sbabu-salesforce: looks like our backups are already restoring with these reserved columns. Whew!

Makefile
```diff
@@ -14,8 +14,5 @@ integration:
 validate:
 	mvn checkstyle:check@checkstyle-execution dependency:analyze@analyze
 
-codecov:
-    mvn cobertura:cobertura
```
sbabu-salesforce: there were spaces here not tabs, so this make target has been failing on travis ci since it was introduced

PR #611, opened 2019-10-21
Logical partitioning: support for updates and stream records

hhkwong:
The main purpose is to support updates and stream records when using logical/hash partitioning. Also, encoding of context + virtualTableName + bucket in the physical hash key has been changed to be length-based so that we can potentially use non-string contexts.

Notes:
- For mapping update expressions and query key condition expressions, both hash and random partitioning now share largely the same logic for parsing and processing. As a result, random partitioning now supports BETWEEN in query range key conditions. Random partitioning's filter expression mapping is still using the existing logic, since it's a no-op for hash partitioning.
- The test ArgumentBuilder now includes the hash partitioning strategy as well. There is some refactoring to make it easy to skip a particular strategy, because there is one test case in UpdateTest that we want to skip for hash partitioning, since it updates only one but not all fields of a secondary index.
- I think a lot of streaming tests still use random partitioning since that's the default in the shared table builder. I didn't change all of them. I just added hash partitioning as one of the strategies in MtAmazonDynamoDbStreamsBaseTest.


src/main/java/com/salesforce/dynamodbv2/mt/mappers/sharedtable/impl/AbstractConditionMapper.java
```diff
@@ -0,0 +1,355 @@
+/*
```
hhkwong: The key condition mapping logic is basically the same as what we had in HashPartitioningConditionMapper before. The update expression mapping logic is new. Both strategies now share this base class.

src/main/java/com/salesforce/dynamodbv2/mt/mappers/sharedtable/impl/HashPartitioningConditionMapper.java
```diff
@@ -7,55 +7,75 @@
...
+        }
+        return fieldPartners;
     }
 
     @Override
-    public void applyForUpdate(UpdateItemRequest updateItemRequest) {
-        // TODO
+    protected void validateFieldsCanBeUpdated(Set<String> allSetFields) {
+        super.validateFieldsCanBeUpdated(allSetFields);
+
```
hhkwong: This is where we enforce that a user can't update just one but not both of the fields in a composite secondary index

src/main/java/com/salesforce/dynamodbv2/mt/mappers/sharedtable/impl/HashPartitioningItemMapper.java
```diff
@@ -53,7 +54,7 @@
         Map<String, AttributeValue> physicalItem = new HashMap<>(virtualItem);
 
         // copy table primary key fields
-        addPhysicalKeys(virtualItem, physicalItem, virtualTable.getPrimaryKey(), physicalTable.getPrimaryKey(), true);
+        addPhysicalKeys(virtualItem, physicalItem, virtualTable.getPrimaryKey(), physicalTable.getPrimaryKey(), false);
```
hhkwong: This is because I want to be able to use this to convert the field-value map represented in an update expression. For put requests, now without this validation, if the item is missing a table PK field, then the mapped item will be missing the physical HK and RK fields, and when we do the put on the parent dynamo client, it will fail. (Adding a test for this now.)

src/main/java/com/salesforce/dynamodbv2/mt/mappers/sharedtable/impl/HashPartitioningKeyMapper.java
```diff
@@ -26,30 +28,33 @@
...
-    @VisibleForTesting
-    static final String DELIMITER = ".";
-
-    private static final String DELIMITER_PATTERN = "\\.";
-
     static class HashPartitioningKeyPrefixFunction {
 
+        // see https://docs.aws.amazon.com/amazondynamodb/latest/developerguide/Limits.html#limits-partition-sort-keys
+        private static final int MAX_HASH_KEY_LENGTH = 2048;
+
```
hhkwong: As discussed in the previous PR, using length-based encoding so that we can easily use non-string contexts in the future if we want.

src/test/java/com/salesforce/dynamodbv2/UpdateTest.java
```diff
@@ -233,26 +237,30 @@ private void runUpdateConditionalOnGsiRkTest(TestArgument testArgument, boolean
                 .withKey(ItemBuilder.builder(testArgument.getHashKeyAttrType(), HASH_KEY_VALUE)
                     .rangeKey(S, RANGE_KEY_OTHER_S_VALUE).build())
                 .addExpressionAttributeValuesEntry(
-                    ":newValue", createStringAttribute(INDEX_FIELD_VALUE + TABLE3 + org + "Updated"))
+                    ":hkValue", createAttributeValue(S, GSI2_HK_FIELD_VALUE + TABLE3 + org + "Updated"))
```
hhkwong: Tweaked a couple of test cases so that we don't modify just one but not all fields of a secondary index, without changing what the test case is essentially testing. And then disabled one test case for hash partitioning, where the scenario being tested requires not updating all secondary index fields at once.

src/main/java/com/salesforce/dynamodbv2/mt/mappers/sharedtable/impl/AbstractConditionMapper.java
```diff
@@ -0,0 +1,355 @@
...
+            Map<String, AttributeValue> virtualSetActions = parseUpdateExpression(request,
+                updateItemRequest.getConditionExpression());
+            checkArgument(!virtualSetActions.isEmpty(), "Update expression needs at least one SET action");
+
+            // validate no table primary key field is being updated, and that if one field in a secondary index is
+            // being updated, then the other is as well
+            validateFieldsCanBeUpdated(virtualSetActions.keySet());
+
+            // map virtual field values to physical ones
+            Map<String, AttributeValue> physicalSetActions = itemMapper.applyForWrite(virtualSetActions);
```
busjaeger: Is this where we will fail if an update is missing all index fields?

busjaeger: Never mind, just saw the comment on the previous line :)

PR #613, opened 2019-10-22
Perf optimizations

busjaeger:


src/main/java/com/salesforce/dynamodbv2/mt/mappers/sharedtable/impl/TableMappingFactory.java
```diff
@@ -80,7 +76,7 @@ public TableMappingFactory(CreateTableRequestFactory createTableRequestFactory,
         this.partitioningStrategy = partitioningStrategy;
         this.virtualTableCreationValidator = new VirtualTableCreationValidator(partitioningStrategy);
         this.pollIntervalSeconds = pollIntervalSeconds;
-        this.physicalTableDescriptionCache = CacheBuilder.newBuilder().maximumSize(100).build();
+        this.physicalTableDescriptions = new ConcurrentHashMap<>();
```
busjaeger: Assumes we have a fixed number of immutable physical tables

PR #620, opened 2019-10-25
replace String equals enum with String equals String

horwitzj:


src/main/kotlin/com/salesforce/dynamodbv2/mt/backups/MtBackupTableSnapshotter.kt
```diff
@@ -66,11 +66,11 @@ open class MtBackupTableSnapshotter {
         do {
             backupStatus = snapshotRequest.amazonDynamoDb.describeBackup(
                     DescribeBackupRequest().withBackupArn(backupResult.backupDetails.backupArn))
-            if (!backupStatus.backupDescription.backupDetails.backupStatus.equals(BackupStatus.CREATING)) {
+            if (backupStatus.backupDescription.backupDetails.backupStatus != BackupStatus.CREATING.toString()) {
```
horwitzj: I have no idea why AWS decided that BackUpDetails.getBackupStatus would return a String instead of an enum

PR #634, opened 2019-11-05
Add additional record info to mt record result

busjaeger:


src/main/java/com/salesforce/dynamodbv2/mt/mappers/MtAmazonDynamoDbStreams.java
```diff
@@ -23,10 +25,60 @@
     class MtGetRecordsResult extends GetRecordsResult {
```
busjaeger: I plan to use these additional attributes to emit metrics / log lines in ZOS.

src/main/java/com/salesforce/dynamodbv2/mt/mappers/MtAmazonDynamoDbStreams.java
```diff
@@ -40,6 +92,32 @@ public MtGetRecordsResult withLastSequenceNumber(String lastSequenceNumber) {
...
+        public MtGetRecordsResult withLastApproximateCreationDateTime(Date lastApproximateCreationDateTime) {
+            setLastApproximateCreationDateTime(lastApproximateCreationDateTime);
+            return this;
+        }
+
+        public Integer getRecordCount() {
+            return recordCount;
+        }
+
+        public void setRecordCount(Integer recordCount) {
```
sbabu-salesforce: how is this different than `records.size()`, ie: can this value be derived?

busjaeger: This is the number of records loaded from the underlying stream before filtering them down to tenant records.

src/main/java/com/salesforce/dynamodbv2/mt/mappers/sharedtable/impl/MtAmazonDynamoDbStreamsBySharedTable.java
```diff
@@ -104,117 +106,104 @@ protected MtGetRecordsResult getAllRecords(GetRecordsRequest request, StreamArn
     @Override
     protected MtGetRecordsResult getRecords(GetRecordsRequest request, MtStreamArn mtStreamArn) {
```
busjaeger: This method changed a bit more than expected, because adding the additional properties to the mt result allowed me to simplify it a lot.

src/main/java/com/salesforce/dynamodbv2/mt/mappers/MtAmazonDynamoDbStreams.java
sbabu-salesforce: should we call this `scannedRecordCount` maybe?

busjaeger: That's a good idea. Will change.

busjaeger: Should I rename the other properties too ... e.g., `firstScannedSequenceNumber` etc.? Or maybe create a separate contained class `ScanResult` or something that contains these five properties?

sbabu-salesforce: keeping this instrumentation in a `ScannedRange` sounds good

busjaeger: @sbabu-salesforce addressed in latest commit. Actually made the code much cleaner overall; thanks for the feedback! Please take a look.

PR #635, opened 2019-11-06
Add composite flavor of MtAmazonDynamoDb

hhkwong:


src/main/java/com/salesforce/dynamodbv2/mt/mappers/MtAmazonDynamoDbComposite.java
```diff
@@ -0,0 +1,169 @@
...
+ */
+public class MtAmazonDynamoDbComposite extends MtAmazonDynamoDbBase {
+
+    private final Collection<MtAmazonDynamoDbBase> delegates;
+    private final Supplier<MtAmazonDynamoDbBase> fromContextToDelegate;
+    private final Function<String, MtAmazonDynamoDbBase> fromPhysicalTableNameToDelegate;
+
+    public MtAmazonDynamoDbComposite(Collection<MtAmazonDynamoDbBase> delegates,
+                                     Supplier<MtAmazonDynamoDbBase> fromContextToDelegate,
+                                     Function<String, MtAmazonDynamoDbBase> fromPhysicalTableNameToDelegate) {
```
hhkwong: Basically the only difference from the previous PR is here, that instead of explicitly passing in the parent MtAmazonDynamoDbBase arguments like the mt context provider, validate all delegates share the same such arguments, and then use the first delegate's properties.

src/test/java/com/salesforce/dynamodbv2/testsupport/ArgumentBuilder.java
```diff
@@ -81,14 +84,14 @@
             .withBinaryHashKey(true)
             .build()),
 
-        HashPartitioning(root -> SharedTableBuilder.builder()
+        HashPartitioning(root -> getTrivialCompositeClient(SharedTableBuilder.builder()
```
hhkwong: In the interest of not adding significantly to the test runtime, I'm making the hash partitioning test argument a composite client of one -- so we're testing both here.

src/test/java/com/salesforce/dynamodbv2/mt/mappers/MtAmazonDynamoDbCompositeTest.java
```diff
@@ -0,0 +1,77 @@
...
+import com.salesforce.dynamodbv2.mt.context.MtAmazonDynamoDbContextProvider;
+import io.micrometer.core.instrument.MeterRegistry;
+import java.util.Collection;
+import java.util.Collections;
+import java.util.stream.Stream;
+import org.junit.jupiter.params.ParameterizedTest;
+import org.junit.jupiter.params.provider.Arguments;
+import org.junit.jupiter.params.provider.MethodSource;
+
+class MtAmazonDynamoDbCompositeTest {
```
sbabu-salesforce: I think a happy path test case demonstrating the toggle `fromContextToDelegate` appropriately sends data to the different delegates, maybe validating on a multi tenant scan across the two shared tables would be nice

src/main/java/com/salesforce/dynamodbv2/mt/mappers/MtAmazonDynamoDbComposite.java
```diff
@@ -0,0 +1,169 @@
...
+    }
+
+    private static MtAmazonDynamoDbBase validateDelegatesAndGetFirst(Collection<MtAmazonDynamoDbBase> delegates) {
+        checkArgument(delegates != null && !delegates.isEmpty(), "Must provide at least one delegate");
+        Iterator<MtAmazonDynamoDbBase> iterator = delegates.iterator();
+        MtAmazonDynamoDbBase first = iterator.next();
+        while (iterator.hasNext()) {
+            MtAmazonDynamoDbBase next = iterator.next();
+            checkArgument(Objects.equals(first.getMtContext(), next.getMtContext()),
+                "Delegates must share the same mt context provider");
```
busjaeger: So if we used different instances of the same context provider, we'd have to make sure to override `equals`?

hhkwong: Yeah we'd have to do that. Right now with the DynamoDbConfig changes I propose in the other PR, both the random partitioning client and the hash partitioning client have the same instance of MtAmazonDynamoDbContextProvider.

sbabu-salesforce:
looks like the code coverage tool is not happy with the coverage against changes in `MtAmazonDynamoDbStreams` and `MtAmazonDynamoDbStreamsComposite`, a unit test on that couldn't hurt

PR #648, opened 2019-11-15
Create and delete virtual multitenant tables

hhkwong:


src/main/java/com/salesforce/dynamodbv2/mt/mappers/MtAmazonDynamoDbBase.java
```diff
@@ -149,15 +149,20 @@ protected MtAmazonDynamoDbContextProvider getMtContext() {
...
     /**
-     * Determines if the table for the given name is a multitenant table associated with this instance.
+     * Determines if the table for the given name is a physical table associated with this instance.
      *
      * @param tableName Name of the table.
-     * @return true if the given table name is a multitenant table associated with this instance, false otherwise.
+     * @return true if the given table name is a physical table associated with this instance, false otherwise.
      */
-    protected boolean isMtTable(String tableName) {
+    protected boolean isPhysicalTable(String tableName) {
```
hhkwong: To make this clear that by this we mean a physical table associated with the instance, as opposed to a virtual multitenant table.

src/main/java/com/salesforce/dynamodbv2/mt/mappers/sharedtable/CreateTableRequestFactory.java
```diff
@@ -21,14 +21,26 @@
...
+     * Returns a list of CreateTableRequests for the static physical tables created when the factory is initialized.
      */
-    List<CreateTableRequest> getPhysicalTables();
+    List<CreateTableRequest> getStaticPhysicalTables();
+
+    /**
+     * Takes a virtual table description and returns a CreateTableRequest for the corresponding physical table to be
+     * created dynamically.
+     */
+    CreateTableRequest getDynamicPhysicalTable(DynamoTableDescription virtualTableDescription);
```
hhkwong: This tells us how to create a new physical table on the fly when there's a request to create a new multitenant table

src/main/java/com/salesforce/dynamodbv2/mt/mappers/sharedtable/CreateTableRequestFactory.java
```diff
@@ -21,14 +21,26 @@
...
+
+    /**
+     * Returns whether the given table name is of a physical table (static or dynamic) belonging to this factory.
+     */
+    boolean isPhysicalTable(String tableName);
```
hhkwong: This is based on table prefix, with the assumption that each mt-dynamo client uses a unique table prefix. Before, we know the exact set of physical tables because we have only static tables. But now we have tables created dynamically / on the fly, so it becomes tricky for each server to keep up-to-date with the current set of physical tables.

src/main/java/com/salesforce/dynamodbv2/mt/mappers/sharedtable/SharedTableBuilder.java
```diff
@@ -318,24 +322,28 @@ public MtAmazonDynamoDbBySharedTable build() {
         setDefaults();
         withName("SharedTableBuilder");
         validate();
+        PhysicalTableManager physicalTableManager = new PhysicalTableManager(amazonDynamoDb, pollIntervalSeconds);
```
hhkwong: Abstracted out physical table create/describe logic from TableMappingFactory into the new PhysicalTableManager

src/main/java/com/salesforce/dynamodbv2/mt/mappers/sharedtable/SharedTableBuilder.java
```diff
@@ -318,24 +322,28 @@ public MtAmazonDynamoDbBySharedTable build() {
...
         if (tableMappingFactory == null) {
             CreateTableRequestFactory createTableRequestFactory = new SharedTableCreateTableRequestFactory(
-                partitioningStrategy.getTablePrimaryKeyMapper(), createTableRequests, getTablePrefix());
+                partitioningStrategy.getTablePrimaryKeyMapper(), createTableRequests,
+                partitioningStrategy::toCompatiblePhysicalPrimaryKey, tablePrefix,
```
hhkwong: TablePartitioningStrategy now has a new method toCompatiblePhysicalPrimaryKey() that converts a given virtual PK  into a physical PK, so that when there's a request to create a new multitenant table, we know how to convert the virtual table CreateTableRequest into a physical one.

src/main/java/com/salesforce/dynamodbv2/mt/mappers/sharedtable/SharedTableBuilder.java
```diff
@@ -692,10 +720,31 @@ private boolean isEmpty(List<?> l) {
...
 
         @Override
-        public List<CreateTableRequest> getPhysicalTables() {
-            return createTableRequests;
+        public List<CreateTableRequest> getStaticPhysicalTables() {
+            return staticPhysicalTables;
+        }
+
+        @Override
+        public CreateTableRequest getDynamicPhysicalTable(DynamoTableDescription virtualTable) {
```
hhkwong: This is called when we want to create a new multitenant table, which means we create a new physical table dynamically. To do this, we convert each table or secondary index PK of the virtual table into a physical PK, according to how the TablePartitioningStrategy defines the conversion.

src/main/java/com/salesforce/dynamodbv2/mt/mappers/sharedtable/impl/TableMappingFactory.java
```diff
@@ -44,55 +40,67 @@
...
     }
 
-    void validateCreateVirtualTableRequest(CreateTableRequest createVirtualTableRequest) {
-        DynamoTableDescription virtualTable = new DynamoTableDescriptionImpl(createVirtualTableRequest);
-        DynamoTableDescription physicalTable = lookupPhysicalTable(virtualTable);
+    /**
+     * Creates the table mapping for a new virtual table, creating the physical table if it doesn't exist.
+     *
+     */
+    TableMapping createNewTableMapping(CreateTableRequest virtualCreateTableRequest, boolean isMultitenant) {
```
hhkwong: Now in addition to the existing getTableMapping(), we also have createNewTableMapping(). createNewTableMapping() is called when MtAmazonDynamoDbBySharedTable.createTable() or createMultitenantTable() is called, and creates the corresponding physical table if needed. getTableMapping() used to create the physical table if it doesn't exist, but now it doesn't -- it only calls describe.

hhkwong: Upon createTable, we used to only validate the virtual and physical table definitions, without creating the physical table if it doesn't exist. Now we make sure the physical table is created right away, instead of possibly waiting until the first DML.

src/main/java/com/salesforce/dynamodbv2/mt/mappers/sharedtable/impl/TableMappingFactory.java
```diff
@@ -101,17 +109,33 @@ void validateCreateVirtualTableRequest(CreateTableRequest createVirtualTableRequ
...
      * corresponding physical table.  Throws a ResourceNotFoundException if the implementation returns null.
      */
-    private DynamoTableDescription lookupPhysicalTable(DynamoTableDescription virtualTable) {
-        return new DynamoTableDescriptionImpl(
-            createTableRequestFactory.getCreateTableRequest(virtualTable).orElseThrow(() ->
-                new ResourceNotFoundException("table " + virtualTable.getTableName() + " is not a supported table")));
+    private CreateTableRequest lookupPhysicalTable(VirtualDynamoTableDescription virtualTable) {
+        // if this is a multitenant table, then create a new physical table dedicated to this purpose.
+        // otherwise, find the corresponding static physical table.
+        return virtualTable.isMultitenant()
```
hhkwong: If this is a multitenant table, then create a dedicated physical table for it on the fly. Thought about making this not hard-coded here, and letting CreateTableRequestFactory decide whether to use one of the static tables or create a new physical table, but if we make it more flexible, then deleteTable becomes trickier (if the MT table doesn't necessarily have its own physical table, then we can't just delete the physical table).

src/main/java/com/salesforce/dynamodbv2/mt/mappers/sharedtable/SharedTableBuilder.java
```diff
@@ -189,6 +192,7 @@
     private String name;
     private AmazonDynamoDB amazonDynamoDb;
     private MtAmazonDynamoDbContextProvider mtContext;
+    private Optional<String> topLevelContext = empty();
```
hhkwong: A shared table client must define a top-level context to be able to create multitenant tables. Creating and deleting multitenant tables can be done only when the context matches the top-level context.

src/main/java/com/salesforce/dynamodbv2/mt/mappers/sharedtable/impl/MtAmazonDynamoDbBySharedTable.java
```diff
@@ -195,16 +197,18 @@ Clock getClock() {
     }
 
     @Override
-    protected boolean isMtTable(String tableName) {
-        return mtTables.containsKey(tableName) && !tableName.startsWith(backupTablePrefix);
+    protected boolean isPhysicalTable(String tableName) {
```
hhkwong: Again, this is now based on table prefix, instead of looking at the set of static physical tables, because now we have physical tables created on the fly. Need to exclude the table description repo table because that also has the same prefix but isn't a data table.

src/main/java/com/salesforce/dynamodbv2/mt/mappers/sharedtable/impl/MtAmazonDynamoDbBySharedTable.java
```diff
@@ -166,15 +169,14 @@ public MtAmazonDynamoDbBySharedTable(String name,
...
         this.name = name;
+        this.topLevelContext = topLevelContext;
         this.mtTableDescriptionRepo = mtTableDescriptionRepo;
         this.tableMappingCache = new MtCache<>(mtContext, tableMappingCache);
         this.tableMappingFactory = tableMappingFactory;
         this.partitioningStrategy = partitioningStrategy;
+        this.physicalTableManager = physicalTableManager;
         this.deleteTableAsync = deleteTableAsync;
         this.truncateOnDeleteTable = truncateOnDeleteTable;
-        this.mtTables = tableMappingFactory.getCreateTableRequestFactory().getPhysicalTables().stream()
```
hhkwong: I think we don't need this because PhysicalTableManager already has a cache of the physical table descriptions that this server knows about (and that cache used to be in TableMappingFactory)

src/main/java/com/salesforce/dynamodbv2/mt/mappers/sharedtable/impl/PhysicalTableManager.java
```diff
@@ -0,0 +1,60 @@
...
+import com.salesforce.dynamodbv2.mt.admin.AmazonDynamoDbAdminUtils;
+import com.salesforce.dynamodbv2.mt.mappers.metadata.DynamoTableDescription;
+import com.salesforce.dynamodbv2.mt.mappers.metadata.DynamoTableDescriptionImpl;
+import java.util.Map;
+import java.util.concurrent.ConcurrentHashMap;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+public class PhysicalTableManager {
+
```
hhkwong: Moved the create & describe table logic plus the physical table description cache from TableMappingFactory to here. Also have a delete table method.

src/main/java/com/salesforce/dynamodbv2/mt/mappers/sharedtable/impl/MtAmazonDynamoDbBySharedTable.java
```diff
@@ -301,10 +317,16 @@ private static void validateGetItemKeysAndAttribute(KeysAndAttributes keysAndAtt
...
     public CreateTableResult createTable(CreateTableRequest createTableRequest) {
-        tableMappingFactory.validateCreateVirtualTableRequest(createTableRequest);
-        // TODO: persist virtual to physical secondary index map in repo
-        return new CreateTableResult()
-            .withTableDescription(withTenantStreamArn(mtTableDescriptionRepo.createTable(createTableRequest)));
+        return createTable(createTableRequest, false);
+    }
+
+    private CreateTableResult createTable(CreateTableRequest createTableRequest, boolean isMultitenant) {
+        TableMapping tableMapping = tableMappingFactory.createNewTableMapping(createTableRequest, isMultitenant);
```
hhkwong: TableMappingFactory.createNewTableMapping() validates the request and creates the physical table if it doesn't already exist.

src/main/java/com/salesforce/dynamodbv2/mt/mappers/sharedtable/impl/MtAmazonDynamoDbBySharedTable.java
```diff
@@ -830,44 +855,50 @@ public String toString() {
...
-                ScanResult scanResult = scan(scanRequest);
-                log.warn("truncating " + scanResult.getItems().size() + " items from table=" + tableName);
-                for (Map<String, AttributeValue> item : scanResult.getItems()) {
-                    deleteItem(new DeleteItemRequest().withTableName(tableName)
-                        .withKey(getKeyFromItem(item, tableName)));
-                }
-                deletedCount += scanResult.getItems().size();
-                if (scanResult.getLastEvaluatedKey() == null) {
-                    break;
+            if (tableDescription.isMultitenant()) {
```
hhkwong: If this is a multitenant table, then drop the corresponding physical table

src/main/java/com/salesforce/dynamodbv2/mt/mappers/metadata/VirtualDynamoTableDescriptionImpl.java
```diff
@@ -0,0 +1,52 @@
...
+    implements VirtualDynamoTableDescription {
+
+    private final boolean isMultitenant;
+
+    public VirtualDynamoTableDescriptionImpl(CreateTableRequest createTableRequest, boolean isMultitenant) {
+        super(createTableRequest);
+        this.isMultitenant = isMultitenant;
+    }
+
+    public VirtualDynamoTableDescriptionImpl(MtTableDescription tableDescription) {
```
hhkwong: MtTableDescription extends TableDescription, with an extra flag `isMultitenant` (and more properties in the future), and that's what we store in the table description repo now. This should make us still be able to parse existing TableDescription records.

src/main/java/com/salesforce/dynamodbv2/mt/repo/MtDynamoDbTableDescriptionRepo.java
```diff
@@ -139,41 +135,48 @@ private TableDescription getTableDescriptionFromCache(String tableName) throws R
         }
     }
 
-    private TableDescription getTableDescriptionNoCache(String tableName) {
+    private MtTableDescription getTableDescriptionNoCache(String tableName) {
```
hhkwong: This is where we first try to get a description for the given virtual table name at the current context, and if it doesn't exist, see if there's virtual multitenant table with this name at the top-level context.

src/main/java/com/salesforce/dynamodbv2/mt/mappers/sharedtable/impl/MtAmazonDynamoDbBySharedTable.java
busjaeger: Is this filter going to match the `*.Leases` tables?

src/main/java/com/salesforce/dynamodbv2/mt/mappers/sharedtable/SharedTableBuilder.java
```diff
@@ -692,10 +720,31 @@ private boolean isEmpty(List<?> l) {
...
         @Override
-        public List<CreateTableRequest> getPhysicalTables() {
-            return createTableRequests;
+        public List<CreateTableRequest> getStaticPhysicalTables() {
+            return staticPhysicalTables;
+        }
+
+        @Override
+        public CreateTableRequest getDynamicPhysicalTable(DynamoTableDescription virtualTable) {
+            String physicalTableName = prefix(tablePrefix, virtualTable.getTableName());
```
busjaeger: Do we need/want to enforce namespacing between the static physical tables configured on the adapter and the dynamic physical tables created for virtual mt tables? If not, what would happen if we create a virtual table with the same name as a static physical table?

src/main/java/com/salesforce/dynamodbv2/mt/mappers/sharedtable/impl/MtAmazonDynamoDbBySharedTable.java
```diff
@@ -115,14 +116,14 @@
     private static final Logger log = LoggerFactory.getLogger(MtAmazonDynamoDbBySharedTable.class);
 
     private final String name;
-
+    private final Optional<String> topLevelContext;
```
busjaeger: What is the purpose of this context?

src/main/java/com/salesforce/dynamodbv2/mt/mappers/sharedtable/impl/TableMappingFactory.java
```diff
@@ -101,17 +109,33 @@ void validateCreateVirtualTableRequest(CreateTableRequest createVirtualTableRequ
...
         // validate secondary indexes
-        virtualTableCreationValidator.validateAndGetSecondaryIndexMap(virtualTable, physicalTable);
+        Map<DynamoSecondaryIndex, DynamoSecondaryIndex> secondaryIndexMap =
+            virtualTableCreationValidator.validateAndGetSecondaryIndexMap(virtualTable, physicalTable);
+
+        // create the physical table for a new virtual table, or describe the physical table for an existing virtual
+        // table. set the returned physical table description back onto the table mapping, so it includes things that
+        // can only be determined after the physical table is created, like the streamArn.
+        physicalTable = createPhysicalTableIfNotExists
+            ? physicalTableManager.createTableIfNotExists(physicalCreateTableRequest)
```
busjaeger: Are we creating physical tables synchronously here?

src/main/java/com/salesforce/dynamodbv2/mt/repo/MtDynamoDbTableDescriptionRepo.java
```diff
@@ -139,41 +135,48 @@ private TableDescription getTableDescriptionFromCache(String tableName) throws R
...
+        // first get description for table name at the current context
+        Optional<MtTableDescription> tableDescription = getTableDescriptionFromDb(tableName);
```
busjaeger: Do we need/want to namespace virtual mt tables and virtual tables? If not, would tenant-level tables shadow mt tables?

src/main/java/com/salesforce/dynamodbv2/mt/mappers/sharedtable/impl/MtAmazonDynamoDbBySharedTable.java
hhkwong: Oh yeah it is going to match that too... Maybe it'd be better to make dynamically created tables have an extra suffix to the prefix, e.g., [clientTablePrefix].d, or something like that, and look for only the static data tables and dynamic data tables found in this way.

What are the rules for mt-dynamo table names and prefixes? Are they defined somewhere?

src/main/java/com/salesforce/dynamodbv2/mt/mappers/sharedtable/SharedTableBuilder.java
hhkwong: Yes I think that's a good idea. Perhaps dynamic physical tables can have a different prefix, which would make isPhysicalTables() more robust / make it less likely that we accidentally pick up other tables like .leases tables.

src/main/java/com/salesforce/dynamodbv2/mt/mappers/sharedtable/impl/MtAmazonDynamoDbBySharedTable.java
hhkwong: This is so that we know whether we're allowed to create/delete/update a multitenant table or to do a cross-tenant scan on a multitenant table. For example this would be the Core cloud context.

busjaeger: Is this different from "no context" (as used in master for cross-tenant scans or change streaming)?

src/main/java/com/salesforce/dynamodbv2/mt/mappers/sharedtable/impl/TableMappingFactory.java
hhkwong: Yes it's still synchronous

src/main/java/com/salesforce/dynamodbv2/mt/repo/MtDynamoDbTableDescriptionRepo.java
hhkwong: To me it seems perhaps more confusing for a user to be able to define both a virtual mt table and a virtual tenant-level table with the same name, so I went with making them share the same namespace. What do you mean by tenant-level tables shadowing mt tables?

src/main/java/com/salesforce/dynamodbv2/mt/mappers/sharedtable/impl/MtAmazonDynamoDbBySharedTable.java
hhkwong: Yes this is different from "no context". It seems like when for example the Core cloud does DDL on a virtual multitenant table, or does a cross-tenant scan on a virtual mt table, there would still be a context.
But maybe we shouldn't make the distinction -- that would make things simpler.

PR #666, opened 2019-12-06
expose big decimal encoding/decoding

sbabu-salesforce:


src/test/java/com/salesforce/dynamodbv2/mt/mappers/sharedtable/impl/BigDecimalSortedBytesConverterTest.java
```diff
@@ -44,6 +44,7 @@ void testNumericalOrderPreserved() {
             new BigDecimal("12.34567"),
             new BigDecimal("-987.654"),
             new BigDecimal("1000"),
+            new BigDecimal("14"),
```
busjaeger: Maybe also add a couple sequence number strings we have observed in DynamoDB?

PR #667, opened 2019-12-09
downgrade aws sdk version from zos failures during ci tests

sbabu-salesforce:


sbabu-salesforce:
I'm afraid binary searching is the only way I can think of to find the breakage version.. I can spend 10 mins to see if I can narrow it down a bit more

msgroi:
If we open a ticket with AWS, having the specific version that broke it would expedite I suspect.

sbabu-salesforce:
This was proxy shenanigans present on our CI env, where the sdk added code pulling proxy information from the env in `1.11.580`. Adding a NO_PROXY config solved our problems.

PR #690, opened 2020-01-06
misc cleanup

horwitzj:


horwitzj:
13 whitespace characters were removed... so code coverage is *now* insufficient?

PR #706, opened 2020-01-21
misc cleanup

horwitzj:


src/main/kotlin/com/salesforce/dynamodbv2/mt/backups/MtBackupManager.kt
```diff
@@ -130,9 +130,8 @@ data class MtBackupMetadata(
         }
         val tenantTableCount = hashMapOf<TenantTableBackupMetadata, Long>()
         tenantTableCount.putAll(tenantTables)
-        for (tenantTable in newBackupMetadata.tenantTables.keys) {
-            tenantTableCount[tenantTable] = tenantTables.getOrDefault(tenantTable, 0L) +
-                    newBackupMetadata.tenantTables[tenantTable]!!
+        newBackupMetadata.tenantTables.entries.forEach {
```
horwitzj: this is the only nontrivial one—loop on entries (instead of keys) since keys and values are both used (esp. b/c that way we don't need to deal w/calling get on a key—where Kotlin will want to deal with what it thinks are potential nulls (which aren't))

PR #782, opened 2020-04-09
Allow building shared table with custom CreateTableRequestFactory not based on static list of tables

hhkwong:
mt-dynamo part of the ZOS PR

### DIFF
--- a/src/main/java/com/salesforce/dynamodbv2/mt/mappers/sharedtable/CreateTableRequestFactory.java
+++ b/src/main/java/com/salesforce/dynamodbv2/mt/mappers/sharedtable/CreateTableRequestFactory.java
@@ -9,7 +9,6 @@ package com.salesforce.dynamodbv2.mt.mappers.sharedtable;
 
 import com.amazonaws.services.dynamodbv2.model.CreateTableRequest;
 import com.salesforce.dynamodbv2.mt.mappers.metadata.DynamoTableDescription;
-import java.util.List;
 import java.util.Optional;
 
 /**
@@ -21,14 +20,14 @@ import java.util.Optional;
  */
 public interface CreateTableRequestFactory {
 
-    /*
+    /**
      * Takes a virtual table description and returns a CreateTableRequest corresponding physical table if one is found.
      */
     Optional<CreateTableRequest> getCreateTableRequest(DynamoTableDescription virtualTableDescription);
 
-    /*
-     * Returns a list of CreateTableRequests that will be created when the factory is initialized.
+    /**
+     * Returns whether the given physical table name belongs to a table that can be created by this factory.
      */
-    List<CreateTableRequest> getPhysicalTables();
+    boolean isPhysicalTable(String physicalTableName);
 
 }
--- a/src/main/java/com/salesforce/dynamodbv2/mt/mappers/sharedtable/impl/MtAmazonDynamoDbBySharedTable.java
+++ b/src/main/java/com/salesforce/dynamodbv2/mt/mappers/sharedtable/impl/MtAmazonDynamoDbBySharedTable.java
@@ -59,8 +59,8 @@ import com.google.common.annotations.VisibleForTesting;
 import com.google.common.base.Preconditions;
 import com.google.common.cache.Cache;
 import com.google.common.collect.ImmutableMap;
-import com.google.common.collect.ImmutableSet;
 import com.google.common.collect.Iterables;
+import com.salesforce.dynamodbv2.mt.admin.AmazonDynamoDbAdminUtils;
 import com.salesforce.dynamodbv2.mt.backups.MtBackupAwsAdaptorKt;
 import com.salesforce.dynamodbv2.mt.backups.MtBackupException;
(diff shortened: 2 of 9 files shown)


src/main/java/com/salesforce/dynamodbv2/mt/mappers/sharedtable/impl/PhysicalTableManager.java
```diff
@@ -0,0 +1,72 @@
...
+        final String tableName = physicalTable.getTableName();
+        return physicalTableDescriptions.computeIfAbsent(tableName, ignored ->
+            new DynamoTableDescriptionImpl(dynamoDbAdminUtils.describeTableIfExists(tableName)
+                .map(description -> {
+                    LOG.info(format("Using existing physical table %s", tableName));
+                    return description;
+                }).orElseGet(() -> {
+                    LOG.info(format("Creating physical table %s", physicalTable.getTableName()));
+                    dynamoDbAdminUtils.createTableIfNotExists(physicalTable, pollIntervalSeconds);
+                    return dynamoDbAdminUtils.describeTable(tableName);
```
ifedorenko, line 67: Nitpicking: this existed before this PR, but `AmazonDynamoDB#createTable` already returns `TableDescription`, we can use that to avoid extra remote call here.

src/main/java/com/salesforce/dynamodbv2/mt/mappers/sharedtable/impl/MtAmazonDynamoDbBySharedTable.java
```diff
@@ -730,13 +735,15 @@ public CreateBackupResult createBackup(CreateBackupRequest createBackupRequest)
                 "Multitenant backups cannot backup individual tables, table-name arguments are disallowed");
             backupManager.get().createBackup(createBackupRequest);
 
-            ExecutorService executorService = Executors.newFixedThreadPool(mtTables.keySet().size());
+            // list all physical tables belonging to this shared table client
+            Collection<String> origPhysicalTables = new AmazonDynamoDbAdminUtils(getAmazonDynamoDb()).listTables()
+                .stream().filter(this::isPhysicalTable).collect(Collectors.toSet());
```
ifedorenko, line 740: This first lists all physical tables, then filters the list to one physical table underlying this `MtAmazonDynamoDbBySharedTable` instance. Do I read this correctly? 

Larger DynamoDB accounts may have hundreds if not thousands of physical tables, so I wonder if it makes sense to rework this to avoid `AmazonDynamoDbAdminUtils#listTables()` call.

src/main/java/com/salesforce/dynamodbv2/mt/mappers/sharedtable/CreateTableRequestFactory.java
```diff
@@ -21,14 +20,14 @@
...
      */
     Optional<CreateTableRequest> getCreateTableRequest(DynamoTableDescription virtualTableDescription);
 
-    /*
-     * Returns a list of CreateTableRequests that will be created when the factory is initialized.
+    /**
+     * Returns whether the given physical table name belongs to a table that can be created by this factory.
      */
-    List<CreateTableRequest> getPhysicalTables();
+    boolean isPhysicalTable(String physicalTableName);
```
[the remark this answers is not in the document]
hhkwong: This is backwards compatible as far as I can tell. `getPhysicalTables()` was used to (1) tell whether a given physical table belongs to an mt-dynamo client, and (2) eagerly create tables in `TableMappingFactory` if desired. Now if `SharedTableBuilder` is given a static list of `CreateTableRequest`s, then `isPhysicalTable()` returns whether the given physical table name belongs to one of the names of the `CreateTableRequest`s, so the behavior for (1) should stay the same. For (2), now if `SharedTableBuilder` is given a list of `CreateTableRequest`s and tables are to be created eagerly, then we tell `PhysicalTableManager` to eagerly create that list of tables.

I think these behaviors are also covered by existing tests.

src/main/java/com/salesforce/dynamodbv2/mt/mappers/sharedtable/impl/MtAmazonDynamoDbBySharedTable.java
hhkwong: Good point. I already have a version of `AmazonDynamoDbAdminUtils#listTables()` where you pass in a filter predicate, so I'll use that version instead.

src/main/java/com/salesforce/dynamodbv2/mt/mappers/sharedtable/impl/PhysicalTableManager.java
hhkwong: I started making changes for this but it turns out I have to change a lot more code in `AmazonDynamoDbAdminUtils`, so I'm going to leave it as it is for now... (`AmazonDynamoDbAdminUtils#createTableIfNotExists` waits until the table is active, and I think we might have to wait till the table is active before we do a describe on the table to make sure we get the `LatestStreamArn` of the table. The `TableDescription` that's in a `CreateTableResult` might not be sufficient.)

PR #804, opened 2020-05-01
Pass specific context to TableMappingFactory.getTableMapping

ifedorenko:
TableMapping instances are always created and used in the context
of specific tenant and virtual table name. As such, pass
explicit context string to TableMapping constructor (as opposed
to MtAmazonDynamoDbContextProvider) to make the code easier to
comprehend and safer to reuse.

Signed-off-by: Igor Fedorenko <ifedorenko@salesforce.com>


busjaeger:
Does this result in more objects being created or is just switching from thread local to explicit method parameters?

ifedorenko:
I believe this does not change number of objects. All TableMapping instances are cached in `MtAmazonDynamoDbBySharedTable#tableMappingCache` (where key includes context string, conveniently hidden in MtCache).

busjaeger (approved):

msgroi (approved):

PR #811, opened 2020-05-07
misc cleanup

horwitzj:


src/main/java/com/salesforce/dynamodbv2/mt/mappers/sharedtable/impl/SharedTableMappingDelegate.java
```diff
@@ -194,8 +193,6 @@ public BatchGetItemResult batchGetItem(TableMapping mapping, BatchGetItemRequest
         // batch get
         final BatchGetItemResult physicalResult = getDynamoDB()
             .batchGetItem(physicalRequest);
-        Map<String, List<Map<String, AttributeValue>>> qualifiedItemsByTable = physicalResult
```
horwitzj, line 197: (IF says OK to remove (it's not being read))

PR #818, opened 2020-05-13
misc cleanup

horwitzj:


pom.xml
```diff
@@ -398,7 +398,7 @@
                         <licenseHeader>
                             <file>${project.basedir}/license-header</file>
                         </licenseHeader>
-                        <trimTrailingWhitespace />
+                        <trimTrailingWhitespace/>
```
ifedorenko, line 401: not exactly "trailing whitespace". but close enough :-)

PR #841, opened 2020-06-09
support for add when updating items

degaydorus:
Supports ADD action type in UpdateItemExpression's in addition to existing support for SET.  Restricts ADD support for HashPartitioningConditionMapper if the index field is part of the index key.

### DIFF
--- a/src/main/java/com/salesforce/dynamodbv2/mt/mappers/sharedtable/impl/AbstractConditionMapper.java
+++ b/src/main/java/com/salesforce/dynamodbv2/mt/mappers/sharedtable/impl/AbstractConditionMapper.java
@@ -20,7 +20,6 @@ import com.amazonaws.AmazonServiceException;
 import com.amazonaws.services.dynamodbv2.model.AttributeValue;
 import com.amazonaws.services.dynamodbv2.model.ComparisonOperator;
 import com.amazonaws.services.dynamodbv2.model.UpdateItemRequest;
-import com.amazonaws.services.dynamodbv2.xspec.UpdateAction;
 import com.google.common.annotations.VisibleForTesting;
 import com.salesforce.dynamodbv2.grammar.ExpressionsBaseVisitor;
 import com.salesforce.dynamodbv2.grammar.ExpressionsLexer;
@@ -136,6 +135,16 @@ abstract class AbstractConditionMapper implements ConditionMapper {
         }
     }
 
+    protected void validateUpdatedFieldIsNotInIndexKey(String updatedField, DynamoSecondaryIndex index) {
+        PrimaryKey primaryKey = index.getPrimaryKey();
+        if (updatedField.equals(primaryKey.getHashKey()) || (primaryKey.getRangeKey().isPresent()
+            && updatedField.equals(primaryKey.getRangeKey().get()))) {
+            throw new IllegalArgumentException(
+                String.format("Cannot update attribute %s. This secondary index is part of the index key",
+                    updatedField));
+        }
+    }
+
     private String getUpdateClause(RequestWrapper request, String field, AttributeValue value, String delimiter) {
         String fieldPlaceholder = MappingUtils.getNextFieldPlaceholder(request);
         request.putExpressionAttributeName(fieldPlaceholder, field);
@@ -168,11 +177,20 @@ abstract class AbstractConditionMapper implements ConditionMapper {
             this.addActions = new HashMap<>();
         }
 
-        protected Map<String, AttributeValue> getSetActions() {
+        Map<String, AttributeValue> getSetActions() {
             return setActions;
         }
 
-        public Set<String> getMergedKeySet() {
+        Map<String, AttributeValue> getAddActions() {
+            return addActions;
+        }
+
+        @VisibleForTesting
+        void setSetActions(Map<String, AttributeValue> setActions) {
+            this.setActions = setActions;
+        }
+
+        Set<String> getMergedKeySet() {
             Set<String> mergedKeySet = new HashSet<>();
             mergedKeySet.addAll(setActions.keySet());
             mergedKeySet.addAll(addActions.keySet());
--- a/src/main/java/com/salesforce/dynamodbv2/mt/mappers/sharedtable/impl/AbstractConditionMapper.java
+++ b/src/main/java/com/salesforce/dynamodbv2/mt/mappers/sharedtable/impl/AbstractConditionMapper.java
@@ -30,6 +30,7 @@ import com.salesforce.dynamodbv2.grammar.ExpressionsParser.IdContext;
 import com.salesforce.dynamodbv2.grammar.ExpressionsParser.KeyConditionContext;
 import com.salesforce.dynamodbv2.grammar.ExpressionsParser.LiteralContext;
 import com.salesforce.dynamodbv2.grammar.ExpressionsParser.SetActionContext;
+import com.salesforce.dynamodbv2.mt.mappers.index.DynamoSecondaryIndex;
 import com.salesforce.dynamodbv2.mt.mappers.metadata.DynamoTableDescription;
 import com.salesforce.dynamodbv2.mt.mappers.metadata.PrimaryKey;
 import com.salesforce.dynamodbv2.mt.mappers.sharedtable.impl.RequestWrapper.AbstractRequestWrapper;
@@ -39,7 +40,6 @@ import java.util.Map;
 import java.util.Optional;
 import java.util.Set;
 import java.util.stream.Collectors;
-import kotlin.Pair;
 import org.antlr.v4.runtime.BailErrorStrategy;
 import org.antlr.v4.runtime.BaseErrorListener;
 import org.antlr.v4.runtime.CharStreams;
@@ -59,13 +59,25 @@ abstract class AbstractConditionMapper implements ConditionMapper {
         this.itemMapper = itemMapper;
     }
 
+    public enum UpdateType {
+        SET("SET"),
+        ADD("ADD");
+
+        private final String value;
+
+        UpdateType(String value) {
+            this.value = value;
+        }
+
+        @Override
+        public String toString() {
+            return this.value;
+        }
+    }
+
     private List<String> getPhysicalUpdateExpressionForAction(RequestWrapper request,
                                                                Map<String, AttributeValue> actions,
                                                                String delimiter) {
-        // validate no table primary key field is being updated, and that if one field in a secondary index is
-        // being updated, then the other is as well
-        validateFieldsCanBeUpdated(actions.keySet());
-
         // map virtual field values to physical ones
         Map<String, AttributeValue> physicalActions = itemMapper.applyForWrite(actions);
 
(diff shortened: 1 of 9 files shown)


src/main/java/com/salesforce/dynamodbv2/mt/mappers/sharedtable/impl/AbstractConditionMapper.java
```diff
@@ -57,28 +59,46 @@
...
+            Map<String, AttributeValue> virtualAddActions = virtualActions.getSecond();
+
+            checkArgument(!virtualSetActions.isEmpty() || !virtualAddActions.isEmpty(),
+                "Update expression needs at least one supported action: SET or ADD");
+
+            List<String> setClauses = getPhysicalUpdateExpressionForAction(request, virtualSetActions, " = ");
+            List<String> addClauses = getPhysicalUpdateExpressionForAction(request, virtualAddActions, " ");
+
+            String setExpression = setClauses.isEmpty() ? "" : "SET " + String.join(", ", setClauses);
+            String addExpression = addClauses.isEmpty() ? "" : "ADD " + String.join(", ", addClauses);
```
hhkwong, line 98: I don't know if this will work with hash partitioning, if we're updating a field in a secondary index. If it's an index field, then the virtual numerical attribute that we're adding to will have its value encoded in a binary index attribute, so we can't just do a straight ADD.

degaydorus: I see, that's probably why there is validation for SET in Hash Partitioning?

src/test/java/com/salesforce/dynamodbv2/UpdateTest.java
```diff
@@ -89,55 +92,230 @@ void update(TestArgument testArgument) {
...
                 .build();
             UpdateItemRequest updateItemRequest = new UpdateItemRequest()
                 .withTableName(TABLE1)
                 .withKey(updateItemKey)
-                .withUpdateExpression("set #someField = :someValue")
-                .withExpressionAttributeNames(ImmutableMap.of("#someField", HASH_KEY_FIELD))
+                .withUpdateExpression("add #someOtherField :" + SOME_OTHER_FIELD_VALUE)
+                .withExpressionAttributeNames(ImmutableMap.of("#someOtherField","someOtherField"))
+                .withExpressionAttributeValues(ImmutableMap.of(":" + SOME_OTHER_FIELD_VALUE,
+                    createNumberAttribute("1")));
```
hhkwong, line 106: We should also have a case for when there's already an existing value on an N attribute, and then we do an update on the attribute and ADD on top of the existing value.

hhkwong: Also, to make sure this works when we're ADDing to an attribute that's part of a secondary index, we should have a test case like updateGsiFieldAndQueryGsi()

src/main/java/com/salesforce/dynamodbv2/mt/mappers/sharedtable/impl/AbstractConditionMapper.java
hhkwong: Where is this validation?

src/test/java/com/salesforce/dynamodbv2/UpdateTest.java
degaydorus: Yeah these are the test cases I hadn't updated yet after running into the validation mentioned above: runUpdateConditionalOnGsiRkTest, runUpdateHkRkTableTest, updateGsiFieldAndQueryGsi, updateFieldInMultipleGsis.

src/main/java/com/salesforce/dynamodbv2/mt/mappers/sharedtable/impl/AbstractConditionMapper.java
degaydorus: https://github.com/salesforce/mt-dynamo/blob/27e9bde11e1fc7fe4ad99eca9d688bc45de26900/src/main/java/com/salesforce/dynamodbv2/mt/mappers/sharedtable/impl/HashPartitioningConditionMapper.java#L74

src/test/java/com/salesforce/dynamodbv2/UpdateTest.java
```diff
@@ -134,10 +228,38 @@ void updateSetFieldMultipleTimesFails(TestArgument testArgument) {
...
+    void updateAddFieldMultipleTimesFails(TestArgument testArgument) {
+        testArgument.forEachOrgContext(org -> {
+            Map<String, AttributeValue> updateItemKey = ItemBuilder.builder(testArgument.getHashKeyAttrType(),
+                HASH_KEY_VALUE)
+                .build();
+
+            UpdateItemRequest updateItemRequest = new UpdateItemRequest()
+                .withTableName(TABLE1)
+                .withKey(updateItemKey)
+                .withUpdateExpression("add someField :someValue, " + SOME_FIELD + " :someValue")
```
ifedorenko, line 243: Can you use `SOME_FIELD` both times? I had to lookup `SOME_FIELD` definition to confirm it is actually `someField`.

src/main/java/com/salesforce/dynamodbv2/mt/mappers/sharedtable/impl/HashPartitioningConditionMapper.java
```diff
@@ -62,15 +64,23 @@
...
     @Override
-    protected void validateFieldsCanBeUpdated(Set<String> allSetFields) {
-        super.validateFieldsCanBeUpdated(allSetFields);
+    protected void validateFieldsCanBeUpdated(Set<String> allUpdateFields, UpdateType type) {
+        super.validateFieldsCanBeUpdated(allUpdateFields, type);
 
-        for (String field : allSetFields) {
+        for (String field : allUpdateFields) {
+            if (type.equals(UpdateType.ADD)) {
+                secondaryIndexes.stream().forEach(index -> validateUpdatedFieldIsNotInIndexKey(
```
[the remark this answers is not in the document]
degaydorus: It will fail the validation if the index field is part of the hash or range key for the index.  I think what you're referring to is the primary key validation (unrelated to indexes?), which is checked in this method and called on super in this case `validateFieldsCanBeUpdated`
https://github.com/salesforce/mt-dynamo/blob/351beab4d6522b94d4a56e1ba7820d871b1dabfa/src/main/java/com/salesforce/dynamodbv2/mt/mappers/sharedtable/impl/AbstractConditionMapper.java#L135

or are you referring to something else?

src/main/java/com/salesforce/dynamodbv2/mt/mappers/sharedtable/impl/AbstractConditionMapper.java
```diff
@@ -57,36 +59,72 @@
...
+            Map<String, AttributeValue> virtualAddActions = virtualActions.get(1);
+
+            checkArgument(!virtualSetActions.isEmpty() || !virtualAddActions.isEmpty(),
+                "Update expression needs at least one supported action: SET or ADD");
 
             // validate no table primary key field is being updated, and that if one field in a secondary index is
             // being updated, then the other is as well
-            validateFieldsCanBeUpdated(virtualSetActions.keySet());
+            validateFieldsCanBeUpdated(virtualSetActions.keySet(), UpdateType.SET);
+            validateFieldsCanBeUpdated(virtualAddActions.keySet(), UpdateType.ADD);
```
hhkwong, line 110: This validation now seems too restrictive, in that if we have a secondary index on field A and field B, and the update does a SET on A and an ADD on B, this validation will reject the update, when it should be allowed. The reason we have the restriction is that both virtual index field end up being encoded in the same physical attribute, so as long as both fields are part of an update, we should allow it.

src/main/java/com/salesforce/dynamodbv2/mt/mappers/sharedtable/impl/AbstractConditionMapper.java
```diff
@@ -101,26 +139,37 @@ private void validateUpdatedFieldIsNotInPrimaryKey(String updatedField, String p
...
         String valuePlaceholder = MappingUtils.getNextValuePlaceholder(request);
         request.putExpressionAttributeValue(valuePlaceholder, value);
-        return fieldPlaceholder + " = " + valuePlaceholder;
+        return fieldPlaceholder + delimiter + valuePlaceholder;
     }
 
     @VisibleForTesting
-    static Map<String, AttributeValue> parseUpdateExpression(RequestWrapper request,
-                                                             String conditionExpression) {
+    static List<? extends Map<String, AttributeValue>> parseUpdateExpression(
```
hhkwong, line 161: It'd be better if the return type is more self-explanatory in what it is, e.g., a class called UpdateActions, with two instances of `Map<String, AttributeValue>`, `setActions` and `addActions`. Otherwise, a comment about what the return value means would be useful.

src/main/java/com/salesforce/dynamodbv2/mt/mappers/sharedtable/impl/AbstractConditionMapper.java
hhkwong: I think we want to pass in all virtualActions into `validateFieldsCanBeUpdated`, without the update type specified, here. And then in the hash partitioning implementation, call super, and then for the set of ADD actions, validate they are not in any secondary index.

src/main/java/com/salesforce/dynamodbv2/mt/mappers/sharedtable/impl/AbstractConditionMapper.java
degaydorus: yeah I was thinking similar even though it would change the usage in other places

src/test/java/com/salesforce/dynamodbv2/UpdateTest.java
```diff
@@ -270,6 +494,180 @@ private void runUpdateConditionalOnGsiRkTest(TestArgument testArgument, boolean
         });
     }
 
+    @ParameterizedTest(name = "{arguments}")
+    @ArgumentsSource(DefaultArgumentProvider.class)
+    @DefaultArgumentProviderConfig(tables = { TABLE3 })
+    void updateWithAddConditionalOnGsiRkSuccess_usePlaceholder(TestArgument testArgument) {
+        runUpdateWithAddConditionalOnGsiRkTest(testArgument, true);
+    }
```
hhkwong, line 502: Are these 4 tests really just validating that secondary index fields can't be updated with ADD? If so it'd be better to do a simpler test and merge these.

src/main/java/com/salesforce/dynamodbv2/mt/mappers/sharedtable/impl/AbstractConditionMapper.java
```diff
@@ -101,26 +135,67 @@ private void validateUpdatedFieldIsNotInPrimaryKey(String updatedField, String p
         }
     }
 
-    private String getUpdateSetClause(RequestWrapper request, String field, AttributeValue value) {
+    protected void validateUpdatedFieldIsNotInIndexKey(String updatedField, DynamoSecondaryIndex index) {
+        PrimaryKey primaryKey = index.getPrimaryKey();
+        if (updatedField.equals(primaryKey.getHashKey()) || (primaryKey.getRangeKey().isPresent()
+            && updatedField.equals(primaryKey.getRangeKey().get()))) {
+            throw new IllegalArgumentException(
+                String.format("Cannot update attribute %s. This secondary index is part of the index key",
```
hhkwong, line 127: Nitpick: would be better if error message specifies that this is for ADD only. e.g., "Cannot update attribute $s with ADD"

PR #869, opened 2020-07-07
misc cleanup (@W-7788265)

horwitzj:


src/test/java/com/salesforce/dynamodbv2/grammar/ExpressionsTest.java
```diff
@@ -101,14 +101,14 @@ void testParseUpdateExpression(String expression) {
...
-        for (String ourTreeLevel: ourTreeLevels) {
+        for (String ourTreeLevel : ourTreeLevels) {
             String dynamodbTreeLevel = dynamodbTreeLevels.get(level);
 
             if (dynamodbTreeLevel.contains("_section") || dynamodbTreeLevel.contains("_action")
                 || dynamodbTreeLevel.contains("_value")) {
                 // dynamodb and our tree contain the same text but differ in the delimiter used and case sensitivity
-                dynamodbTreeLevel.replace("_", "").toLowerCase();
-                ourTreeLevel.toLowerCase();
+                assertEquals(dynamodbTreeLevel.replace("_", "").toLowerCase().trim(),
```
horwitzj, line 110: before the change, modifications of `dynamodbTreeLevel` and `ourTreeLevel` were being computed but not tested (unlike in the other (`else if` and `else`) branches)

degaydorus: yes the assertion should be there, thanks

